PR #7, opened 2018-01-25
Add support for optional attributes

aykevl:


requests.go
```diff
@@ -142,21 +142,36 @@ func (cr *CredentialRequest) AttributeList(conf *Configuration) (*AttributeList,
...
+	for crName := range cr.Attributes {
+		found := false
+		for _, ad := range credtype.Attributes {
+			if ad.ID == crName {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return nil, errors.New("Unknown CR attribute")
```
sietseringers: This error message is not clear; it should at least mention the credential type

PR #13, opened 2018-06-08
David pinchange

davidv1992:


irmaclient/client.go
```diff
@@ -683,6 +690,16 @@ func (client *Client) unenrolledSchemeManagers() []irma.SchemeManagerIdentifier
...
 
+func (client *Client) EnrolledSchemeManagers() []irma.SchemeManagerIdentifier {
+	list := []irma.SchemeManagerIdentifier{}
+	for name, manager := range client.Configuration.SchemeManagers {
+		if _, contains := client.keyshareServers[name]; manager.Distributed() && contains {
+			list = append(list, manager.Identifier())
+		}
+	}
+	return list
+}
```
sietseringers: With the exception of an absent `!`, this function is code duplication from `client.unenrolledSchemeManagers()`. Simultaneously there are several inconsistencies between the two functions:
* This one is exported, the other one is not
* The purpose of the other one is to update state (namely the member `client.UnenrolledSchemeManagers`), while the output of this one is used directly in `irma_mobile`.

I think it would be best to solve the code duplication as well as these inconsistencies by:
* Adding a member `EnrolledSchemeManagers` to the `client` immediately below `UnenrolledSchemeManagers`
* Update (and rename) the `client.unenrolledSchemeManagers()` function such that it updates both the `client.EnrolledSchemeManagers` and `client.UnenrolledSchemeManagers` members
* in `irma_mobile`, use the new `client.EnrolledSchemeManagers` member instead of the method `client.EnrolledSchemeManagers()`.

irmaclient/client.go
```diff
@@ -80,10 +80,17 @@ type KeyshareHandler interface {
 	EnrollmentSuccess(manager irma.SchemeManagerIdentifier)
 }
 
+type ChangepinHandler interface {
+	ChangepinFailure(manager irma.SchemeManagerIdentifier, err error)
+	ChangepinSuccess(manager irma.SchemeManagerIdentifier)
+	ChangepinIncorrect(manager irma.SchemeManagerIdentifier)
+}
```
sietseringers: PIN is a separate word from Change, but they are concatenated here, so `ChangePin` would be preferable - here and all other places where `pin` is concatenated with some other word (including`irma_mobile` and `irma_keyshare_server`).

PR #14, opened 2018-06-12
[WIP] Signed JWT support

aykevl:
Work-in-progress PR for verified JWT support in irmago.

Example JWT that I get out of it:
```
eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJQcml2YWN5IGJ5IERlc2lnbiBGb3VuZGF0aW9uIiwiaXNzIjoidmVyaWZpY2F0aW9uX3JlcXVlc3QiLCJpYXQiOjE1Mjg4MzM5MTksInNwcmVxdWVzdCI6eyJ2YWxpZGl0eSI6MTIwLCJyZXF1ZXN0Ijp7ImNvbnRleHQiOm51bGwsIm5vbmNlIjpudWxsLCJjb250ZW50IjpbeyJsYWJlbCI6IkZhbWlseSBuYW1lIiwiYXR0cmlidXRlcyI6WyJwYmRmLnBiZGYuaWRpbi5mYW1pbHluYW1lIl19LHsibGFiZWwiOiJEYXRlIG9mIGJpcnRoIiwiYXR0cmlidXRlcyI6WyJwYmRmLnBiZGYuaWRpbi5kYXRlb2ZiaXJ0aCJdfV19fX0.qHO5wVLdbmuukKWXD1YG_cmSR6mNKV6XbZp7a8MizY5Ve6KXI6HDjkRCUbxRI39SHxzJiYhi8Ar2Lim6ofK0MQxTEVdH4Vp-Lof6VDNvpHZwBZMD1FMz1Lw8jg0Q4WJ5PMpddz9PLf-zFdZCltAOB2RQwOdxkYzrqbyM1Xb57KkVodPKmgs_TFqWG8Ki40oMD6kYUEbbrTl_3VS1cxHZpxlBOqfNWEw4Hk_Kd1qBTLsIR_lwbnA8PyFOgto4Ovjtcft2Oj5vcdXZUePsC3vU1rRCMIPuqjqOtBZ7U1nkYKxiLvfzZC5xqQMfW3avxx8W-uDId5MZ8sxNkTdvqHCKIg
```

### DIFF
--- a/requests.go
+++ b/requests.go
@@ -1,6 +1,7 @@
 package irma
 
 import (
+	"crypto/rsa"
 	"fmt"
 	"math/big"
 	"strconv"
@@ -9,6 +10,7 @@ import (
 	"encoding/json"
 
 	"github.com/bwesterb/go-atum"
+	"github.com/dgrijalva/jwt-go"
 	"github.com/go-errors/errors"
 	"github.com/mhe/gabi"
 )
@@ -82,14 +84,19 @@ type CredentialRequest struct {
 
 // ServerJwt contains standard JWT fields.
 type ServerJwt struct {
-	Type       string    `json:"sub"`
-	ServerName string    `json:"iss"`
-	IssuedAt   Timestamp `json:"iat"`
+	Type      string     `json:"sub"`
+	HumanName string     `json:"iss"`
+	IssuedAt  *Timestamp `json:"iat"`
+}
+
+func (jwt ServerJwt) Valid() error {
+	return errors.New("not implemented")
 }
 
 // A ServiceProviderRequest contains a disclosure request.
 type ServiceProviderRequest struct {
-	Request *DisclosureRequest `json:"request"`
+	Validity int                `json:"validity"`
+	Request  *DisclosureRequest `json:"request"`
 }
 
 // A SignatureRequestorRequest contains a signing request.
@@ -136,6 +143,12 @@ type IrmaSession interface {
 
 // Timestamp is a time.Time that marshals to Unix timestamps.
 type Timestamp time.Time
+
+// Return the current time as a Timestamp value.
+func TimestampNow() *Timestamp {
+	t := time.Now()
+	return (*Timestamp)(&t)
+}
 
 func (cr *CredentialRequest) Info(conf *Configuration, metadataVersion byte) (*CredentialInfo, error) {
 	list, err := cr.AttributeList(conf, metadataVersion)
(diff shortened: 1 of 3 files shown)


jwt.go
```diff
@@ -0,0 +1,25 @@
...
+	"crypto/rsa"
+
+	"github.com/dgrijalva/jwt-go"
+)
+
+func CreateDisclosureJwt(attributes AttributeDisjunctionList, serverName, humanName string, sk *rsa.PrivateKey) (string, error) {
+	claims := ServiceProviderJwt{
+		ServerJwt: ServerJwt{
+			ServerName: serverName,
+			Type:       humanName,
```
sietseringers: * `Type` should (in the case of disclosures) be the fixed string `verification_request`, see `NewServiceProviderJwt()`
* You can actually reuse `NewServiceProviderJwt()` to build the JWT before signing it
* We should have similar functions for signing and issuance (reusing `NewSignatureRequestorJwt()` and `NewIdentityProviderJwt()`

As to the `serverName` and `humanName`, that has now become a little messy:
* The variable `humanName` should be assigned to the field `ServerName`, whose name is inappropriate (I think `JwtIssuer` would be better
* The variable `serverName` should go into the JWT `kid` header.

aykevl: Ah, somehow I managed to miss `NewServiceProviderJwt()`. I've now reused this function.

> We should have similar functions for signing and issuance [...]

Yes, I'll add them later. I would prefer getting this one right first.

> The variable serverName should go into the JWT kid header.

Does this work? At least the Java implementation appears to use `iss`, not `kid` (verified by decoding the JWT for BIG name/birthdate disclosure).

sietseringers: Originally we indeed used only `iss`; this was (1) shown to the user in the IRMA app and (2) the API server used this to lookup the verifier's public key with which to verify its JWTs. At some point we split these into two, so now:
* The API server uses the `kid` aka `serverName` field in the header for the public key lookup
* `iss` aka `humanName` is what is shown to the user in the IRMA app.

For example, here is the header from a decoded IRMATube JWT obtained by clicking on a movie:
```json
{
  "typ": "JWT",
  "alg": "RS256",
  "kid": "irmatube"
}
```
Body:
```json
{
  "iat": 1529269092,
  "iss": "IRMATube",
  "sub": "verification_request",
  "sprequest": { "truncated" }
}
```
These are made by [this PHP library](https://github.com/privacybydesign/irma-requestor/blob/master/src/Requestor.php).

confiks:
Could you please not force push branches, especially when they are tied to pull requests. [It makes it very hard](https://github.com/privacybydesign/irmago/pull/14/files/0ec9427d6086844adc04ba73df56710394fec225) to figure out what the feedback is about. Just add new commits on top of each other.

aykevl:
Ok, will avoid that now. Some projects actually want you to force-push so that's why I did it.

aykevl:
Any update on this? I need this for the new DUO issuer.

requests.go
```diff
@@ -427,3 +443,24 @@ func (jwt *SignatureRequestorJwt) IrmaSession() IrmaSession { return jwt.Request
 
 // IrmaSession returns an IRMA session object.
 func (jwt *IdentityProviderJwt) IrmaSession() IrmaSession { return jwt.Request.Request }
+
+// Sign returns the signed and serialized JWT.
+func (claims *ServiceProviderJwt) Sign(serverName string, sk *rsa.PrivateKey) (string, error) {
```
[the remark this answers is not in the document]
aykevl: I've also updated the other two receivers.

aykevl:
I added the dependency, but avoided the `inputs-digest` line because it resulted in a merge conflict (and according to https://github.com/golang/dep/issues/1224 that's probably safe).

aykevl:
Added a method to verify and parse a disjunction JWT.

aykevl:
Closed in favor of #21.

PR #22, opened 2018-09-27
Added irmago support for optional Disjunctions.

davidv1992:


davidv1992:
No longer relevant given condiscon

PR #42, opened 2019-05-07
Added irmago modifications for forced updates.

davidv1992:


confiks:
@sietseringers Is it okay to call `conf.ParseFolder()` in `UpdateSchemes`. Why didn't this happen before for the auto update mechanism of the server in `AutoUpdateSchemes`?

sietseringers:
Calling `ParseFolders()` is not only okay, it is required; that it wasn't there before was a bug. Any downloaded or updated files would become available to the server only after a restart.

`ParseFolder` is expensive however, so we should call it only when files were actually updated or added. We can keep track of that with the second parameter of `UpdateSchemeManager()`:
```go
func (conf *Configuration) UpdateSchemes() error {
	var updated IrmaIdentifierSet
	for id := range conf.SchemeManagers {
		Logger.WithField("scheme", id).Info("Auto-updating scheme")
		if err := conf.UpdateSchemeManager(id, &updated); err != nil {
			return err
		}
	}
	if updated.Empty() {
		return nil
	}
	return conf.ParseFolder()
}
```

PR #54, opened 2019-06-11
Updated disjunction.attemptSatisfy and disjunction.satisfied to actua…

davidv1992:
…lly compare the values of the attributes, and not just the pointers to those values (which can differ even when values are equal).


davidv1992:
Fixed in condiscon, no longer needed.

PR #56, opened 2019-06-19
Enforce PIN entry when signing

ivard:
Additional case added to make sure that a user's PIN code is always asked when doing a signing session.

### DIFF
--- a/irmaclient/keyshare.go
+++ b/irmaclient/keyshare.go
@@ -197,6 +197,12 @@ func startKeyshareSession(
 		transport.SetHeader(kssVersionHeader, "2")
 		ks.transports[managerID] = transport
 
+		// Enforce that PIN is always asked when signing
+		if session.Action() == "signing" {
+			ks.pinCheck = true
+			continue
+		}
+
 		// Try to parse token as a jwt to see if it is still valid; if so we don't need to ask for the PIN
 		parser := new(jwt.Parser)
 		parser.SkipClaimsValidation = true // We want to verify expiry on our own below so we can add leeway


stevenvegt:
How would this look to the user? I imagine: I open my IRMA app, enter the PIN. Start a signing session and than enter my PIN again, or is the PIN I entered still valid?

ivard:
After the user agrees with signing a message, the user is now always asked to enter is PIN. This goes via the same flow as in the current situation when the previous PIN entry was too long ago (so the PIN session is expired). Indeed when doing everything in line, the user should enter his PIN twice.

It is similar to the approach of banks. There you should also re-enter your PIN when you want to perform a bank transaction to make users aware that they are doing something special now.

stevenvegt:
We at Nuts use a signing session to login in to the application. I believe this extre PIN entry will signifactly reduce the UX. Especially if you consider that the user probably has just entered its pin 5 seconds ago. So, it will not improve security, but reduce UX.
Based on what research or feedback did you decide to implement this change? I would like to ask you to consider our use-case as well. I'm afraid this will not help in the adoption of IRMA.

confiks:
The intention behind the change is to give the user an extra sense of committing to the signature, by asking the PIN again, but I personally quite agree that it will degrade the user experience, especially in your case where you are using a signature session as a means to specify an authentication more clearly.

An option would be to let the verifier choose if an extra step would be necessary. Perhaps you know of a way that will match the intention while impacting the user experience less. We'll discuss it in a next meeting and will get back to you here.

stevenvegt:
Interesting note: its all "client side". There is no crypto enforcing this behaviour. So, as a service provider you may require an extra PIN entry, but a client (app which talks the IRMA protocol) can ignore this request and there is no way of knowing.

ivard:
We know that it is only client side. In the current set-up it is technically not possible to make a difference in this, because the keyshare server is not aware of the exact things the user is doing for privacy reasons. The change would be much more complicated if the keyshare server should also be aware of this. 

Measures like this are of course intended to improve the user's security by preventing he/she does not do anything by accident. It is therefore also in the user's interest to use the right app. Nevertheless, before rolling this out we will first look whether we can improve it for the remarks that you made.

PR #60, opened 2019-07-31
New feature: log screens

ivard:


internal/sessiontest/logs_test.go
```diff
@@ -35,27 +36,34 @@ func TestLogging(t *testing.T) {
 	require.NoError(t, err)
 	require.NotEmpty(t, disclosed)
 
+	time.Sleep(2 * time.Second)
```
davidv1992: Do we need this explicit sleep here? Aren't there better options?

ivard: The sleep is needed to make sure that the next session always falls in the next unix time epoch. Otherwise the test on the before parameter of `LoadLogs` might fail. It would be possible to do it in another way, but then we are testing it in a slightly other way than it is used in production. In production the LoadLogs also "assumes" a unix time in seconds is unique for a session. I added a comment with an explanation in a new commit.

ivard:
I now processed most of the feedback in the latest commit. With testing I however found a problem. In bolthold the uint64 is stored in a string representation. We use the JSON encoder and encoding a uint64 with the JSON encoder gives the string representation of that uint64. This however leads to the issue that the database is not sorted as {1, 2, ..., 10, 11} but as {1, 10, 11, 2, ...}. Therefore the sorting order of the database gets broken.

I tested whether the default encoder Gob could be used, but for `irma.Timestamp` no interface is implemented how to serialize it. Maybe this holds for more data types as `irma.Timestamp`, but I didn't check on that. We have to discuss first how to continue, because this concerns the build-up of the database and we cannot roll out this feature if we are planning to potentially change the bolthold database structure in short time.

I think we have two options. Either we have to change the encoder and decoder in such a way that this issue doesn't occur or we have to do inserting on bbolt directly too. The first scenario can be done by testing how other encoders solve this and whether we can use those or we can make our own decoder that uses JSON for structs and the raw value for single variables. In the second scenario a disadvantage is that we practically skip all functionality from bolthold and use raw bbolt.

PR #64, opened 2019-10-16
Added support for DisallowDelete

ivard:


stevenvegt:
What is a DisallowDelete?

confiks:
`DisallowDelete` is a boolean flag in the `CredentialType` definition in the scheme that (in its proposed form here) disallows deleting the credential. It is intended as a mechanism to make it impossible to delete the 'keyshare server pseudonym' in the mobile application, in the `pbdf` scheme.

Another implementation was proposed by @sietseringers  (on Jul 2):
> Instead of including what you can or cannot *do* with this credential type in the scheme, would it not make more sense to include what this *is*? Namely, the initial attribute after registration. Then the code can its conclusion from that - namely, that this credential is not to be deleted.
> This could then also solve another issue, namely that `irma scheme verify` in the `pbdf` scheme complains that the `mijnirma` credential type has no `IssueURL` tag. If the `irma` tool knows that this is the initial attribute then it can infer that no `IssueURL` tag is required.

Another option would be to take this solely as a UI hint and not make changes to `irmaclient/client.go`.

I think the PR is currently fine as is.

ivard:
When enrolling at IRMA, every user gets his/her MijnIRMA credential, now also known as IRMA pseudonym. This credential can be used for logging in at the management interface MijnIRMA. We saw that some people accidentally deleted their MijnIRMA credential as they sometimes don't understand why they got it. This can be problematic when the user wants to do something with MijnIRMA, because it is not possible to re-gain that credential. Therefore we thought it would be wise if users aren't able to delete this base credential at all. This feature adds the support for a `DisallowDelete` flag in schemes such that we can enable this feature for the pseudonym credential by simply indicating this in the pbdf scheme.

PR #72, opened 2019-11-07
Added port mismatch warning to irma session

ivard:
A mistake I myself often make when using `irma session` is that I change the port in `--url` and then I forget changing the `--port` setting too. There are scenarios thinkable in which you want another port number in the URL than on which the server is running, for example when a proxy is used in between the server and the phone. A difference in port between `--url` and the `--port` should not result in a complete failure. Therefore I would suggest at least warning the user that he is doing something that is probably not right.

Another option would be just using the port from `--url` when `--port` is not specified.


ivard:
Apparantly already possible when using `--url http://localhost:port --port 1234`

PR #84, opened 2020-02-06
Switch to go modules for dependency management

### LINKED ISSUE #78: go mod?
Shouldn't it be wise to migrate to go.mod in the code and the documentation?

I did a go mod init irmago in the irmago folder and it seems that everything works.

Lucus16:
This should be a one to one translation from the `Gopkg.lock` as generated by `go mod init`. I've checked that all local tests still pass.

Closes #78.


sietseringers:
Thanks for the PR, and sorry for reacting so late. I've been waiting with switching to modules for the `revocation` branch to land in `master`, which it just did. (Since a lot has changed in that branch I have redone the conversion.)

PR #101, opened 2020-06-23
Fixed missing '#token=' part in login mail TokenURL

ivard:


davidv1992:
Current behaviour is intended, to keep server more flexible. Intent is that you specify that specific bit in the configuration of the server as well.

PR #102, opened 2020-07-16
Randomblind

leonbotros:
This PR adds support for random blind attributes, enabled for an attribute type by including the `randomblind="true"` XML-attribute in a credential description in the scheme. For example, see [this credential from a test configuration](url). During issuance the issuer does not input a value, but rather a value is constructed from both party's randomness. It is possible for a credential to contain multiple random blind attributes.

Currently, I'm still debating how to handle:
- Attribute values that are yet unknown (e.g., in the credential request, where we are essentially saying "You will be getting some value, but we don't know what exactly"). 
The current state of this PR sends a nil `TranslatedValue` and assumes the app (or any other consumer) will take care of this.
- What attribute value to emit after issuance has been completed. Currently, the PR handles this by decoding to a base-62 string. 
In my opinion, the app should be responsible to convert this into something that is more presentable to the user.

Before merging we should update this PR to use gabi's master branch as a module, once support for random blind issuance has been merged there. See, https://github.com/privacybydesign/gabi/pull/14.


leonbotros:
Thanks for the feedback, I've updated the PR. You are right about not sending over a nil value, I will look into it.

leonbotros:
Yes, you are right on this. I'll add a fix for this.

descriptions.go
```diff
@@ -119,6 +121,33 @@ func (ad AttributeType) IsOptional() bool {
...
+	indices := []int{}
+	for i, at := range ct.AttributeTypes {
+		if at.RandomBlind {
+			indices = append(indices, i+1)
+		}
+	}
+	return indices
+}
+
+func (ct *CredentialType) getAttributeTypeIdentifiers(indices []int) (ids []string) {
```
sietseringers, line 136: Using `get` prefixes for getters is [not idiomatic in Go](https://golang.org/doc/effective_go.html#Getters) (also in `GetRandomBlindAttributeTypeIdentifers`).

irmaclient/session.go
```diff
@@ -326,7 +326,7 @@ func (session *session) processSessionInfo() {
 		issuedAt := time.Now()
 		_, err := ir.GetCredentialInfoList(session.client.Configuration, session.Version, issuedAt)
 		if err != nil {
-			session.fail(&irma.SessionError{ErrorType: irma.ErrorUnknownIdentifier, Err: err})
+			session.fail(err.(*irma.SessionError))
```
sietseringers, line 329: If `err` is not nil then it will not necessarily be of type `*irma.SessionError` (e.g. `requests.go` line 643), so this type assertion can cause a panic. It would be better to check if the error is this type; if so just pass it on; if not wrap it into one.

server/irmaserver/api.go
```diff
@@ -168,8 +168,15 @@ func (s *Server) StartSession(req interface{}, handler server.SessionHandler) (*
 	if err := s.validateRequest(request); err != nil {
 		return nil, "", err
 	}
-
 	if action == irma.ActionIssuing {
+
```
sietseringers, line 172: Looks like you accidentally placed an empty line here, would prefer it to be removed.

sietseringers (changes requested):

PR #103, opened 2020-07-28
Fixed race condition in session dismiss.

davidv1992:


irmaclient/session.go
```diff
@@ -660,13 +668,18 @@ func panicToError(e interface{}) *irma.SessionError {
...
 // returns whether or not it did something.
 func (session *session) finish(delete bool) bool {
-	if !session.done {
-		if delete && session.IsInteractive() {
-			session.transport.Delete()
-		}
-		session.client.nonrevRepopulateCaches(session.request)
-		session.client.sessions.remove(session.token)
-		session.done = true
+	irma.Logger.Warn(fmt.Sprintf("Deleting session %p", session))
```
sietseringers, line 671: - Instead of saying `Logger.Warn(fmt.Sprintf`, you can just say `Logger.Warnf` here
- Better would be to put it in a field: `irma.Logger.WithField("session", session).Warn("...")`
- Please don't use the `Warn` level for non-exceptional situations like these

irmaclient/session.go
```diff
@@ -660,13 +668,18 @@ func panicToError(e interface{}) *irma.SessionError {
...
+	if _, ok := <-session.done; ok {
+		// Do actual delete in background, since that can take a while in some circumstances, and
+		// precise moment of completion isn't relevant for frontend.
+		go func() {
+			if delete && session.IsInteractive() {
+				session.transport.Delete()
+			}
+			session.client.nonrevRepopulateCaches(session.request)
+			session.client.sessions.remove(session.token)
+			irma.Logger.Warn(fmt.Sprintf("Deleted session %p", session))
```
sietseringers, line 681: Same as above.

irmaclient/session.go
```diff
@@ -660,13 +668,18 @@ func panicToError(e interface{}) *irma.SessionError {
...
 func (session *session) finish(delete bool) bool {
-	if !session.done {
-		if delete && session.IsInteractive() {
-			session.transport.Delete()
-		}
-		session.client.nonrevRepopulateCaches(session.request)
-		session.client.sessions.remove(session.token)
-		session.done = true
+	irma.Logger.Warn(fmt.Sprintf("Deleting session %p", session))
+	if _, ok := <-session.done; ok {
```
sietseringers, line 672: This would benefit from a comment explaining what happens the first time and what happens the second (or later) times.

irmaclient/session.go
davidv1992: Woeps, have removed the left-over debugging code

irmaclient/session.go
davidv1992: removed, see above.

irmaclient/session.go
davidv1992: Included an explanation

PR #104, opened 2020-07-31
Added support for pretty verifier names.

davidv1992:
General setup is as follows: If there is a scheme with the name requestor, this is seen as the verified requestor scheme and loaded separately. The information of that scheme is then used in irmaclient to calculate servername based on hostname.

Tooling is added for this under the irma requestor command, where both sign and verify work as expected for the requestor scheme. Irma scheme verify also has awareness of, and validates, the requestor scheme if present


irmaconfig.go
```diff
@@ -59,6 +60,14 @@ type Configuration struct {
 	// (i.e., invalid signature, parsing error), and the problem that occurred when parsing them
 	DisabledSchemeManagers map[SchemeManagerIdentifier]*SchemeManagerError
 
+	// RequestorScheme data of the currently loaded requestorscheme
+	RequestorScheme *RequestorScheme
```
sietseringers, line 64: This supports at most one requestor scheme simultaneously, unlike normal schemes, is that what we want?

irmaconfig.go
```diff
@@ -337,6 +350,111 @@ func (conf *Configuration) ParseSchemeManagerFolder(dir string, manager *SchemeM
...
+	// Verify all files in index, reading the RequestorChunks
+	var requestors []*RequestorInfo
+	for file := range conf.RequestorScheme.index {
+		exists, err = common.PathExists(filepath.Join(conf.Path, file))
+		if err != nil {
+			conf.RequestorScheme.Status = SchemeManagerStatusInvalidSignature
+			return
+		}
+		if !exists {
+			continue
```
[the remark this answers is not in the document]
davidv1992: Good catch, this is also present in the (preexisting) VerifySchemeManager function. Do we also want different behaviour there?

davidv1992:
I have tried to maximize reuse as far as I could see gave reasonable results. If you see further options for code reuse between the two feel free to edit this to do that. There are some parallels in places, but there in my view the equality is mostly structural, and there are enough details different in those places that I don't see merging them as useful.

irmaclient/logs.go
```diff
@@ -35,6 +36,72 @@ type LogEntry struct {
...
+		// Signature sessions
+		SignedMessage          []byte          `json:",omitempty"`
+		Timestamp              *atum.Timestamp `json:",omitempty"`
+		SignedMessageLDContext string          `json:",omitempty"`
+
+		// Issuance sessions
+		IssueCommitment *irma.IssueCommitmentMessage `json:",omitempty"`
+
+		// All session types
+		ServerName *json.RawMessage      `json:",omitempty"`
```
ivard, line 61: Why does this have to be a pointer to json.RawMessage? In `Request` it is not a pointer. When it is not a pointer and it is empty it will just get the value of an empty array and then you can distinguish it too right? Or is the empty message also a specific ServerName?

davidv1992: Per the programmed old format, ServerName can be absent in json logenties. For these cases, the canonical representation with a RequestorInfo is a nilpointer to one, not an empty RequestorInfo. This just ensures that that case can be easily distinguished in the code following.

ivard: Okay, because it is a temporary type it is acceptable then. The full class is still a bit verbose of course (that's why all the manual assignments are needed. Since this is code that cannot be removed at some point (legacy logs will always be there), did you consider parsing this in a `UnmashalJSON` function for `RequestorInfo` instead of doing it in the `UnmarshalJSON` of `LogEntry`? Then you would not need all the manual assignments if I'm right.

davidv1992: Did consider it, but did not go for it as this is where the actual incompatibility arises from, which makes it semantically easier to reason about when fixed here. This also prevents any issues with RequestorInfo being misparsed from schemes, where TranslatedValues are not acceptable.

ivard: So you then mean that this is not the only place where `RequestorInfo` is parsed and at the other places it is actually weird to include the legacy parsing too so if I understand you right?

davidv1992: Yes, RequestorInfo is also parsed directly from the RequestorSchemes, and there legacy parsing is unwanted

ivard: I see, then that's no option indeed. Then it is fine this way I think.

irmaclient/logs.go
```diff
@@ -35,6 +36,72 @@ type LogEntry struct {
...
 }
 
+// We need manual unmarshalling to deal with legacy log entries that have
+func (entry *LogEntry) UnmarshalJSON(b []byte) (err error) {
+	// Internal type needed for unmarshalling to delay requestorInfo unmarshalling
+	// NOTE: keep this up to date with above definition of log entry!
+	var temp struct {
+		// General info
+		ID   uint64
+		Type irma.Action
```
sietseringers, line 46: Instead of this copy of the LogEntry struct that needs to be kept in sync, you can also create a new type with the same underlying type as `LogEntry` by saying
```go
type rawLogEntry LogEntry
var temp rawLogEntry
```

davidv1992: That is not sufficient in this case because I want to delay unmarshaling of the ServerName field, and thus need it to have a different type.

ivard: An option would be to make a type without `ServerName` in it (lets use @sietseringers name `rawLogEntry`) and then make LogEntry `type struct { rawLogEntry, ServerName ... }` by including the other struct. Then you can do the same in your temp struct and include another type for `ServerName`

schemes.go
```diff
@@ -1,70 +1,1151 @@
...
+		}
+		scheme = newScheme(SchemeType(typ))
+		return scheme, common.Unmarshal(filename, b, scheme)
+	}
+
+	return nil, errors.New("no scheme description file found")
+}
+
+func tempSchemeCopy(scheme Scheme) (string, string, error) {
+	dir, err := ioutil.TempDir("", "tempscheme")
```
ivard, line 803: The use of this function causes on Android the error `*os.PathError mkdir /data/local/tmp/tempscheme688734974: permission denied`. Probably Android does not allow apps to write to its default `/tmp` dir.

irmaclient/session.go
```diff
@@ -255,27 +246,20 @@ func (session *session) getSessionInfo() {
...
+func requestorInfo(serverURL string, request irma.SessionRequest, conf *irma.Configuration) *irma.RequestorInfo {
+	if serverURL == "" {
+		return nil
 	}
+	u, _ := url.ParseRequestURI(serverURL) // Qr validator already checked this for errors
+	hostname := u.Hostname()
+	info, present := conf.Requestors[hostname]
 
-	return sn
+	if (u.Scheme == "https" || !common.ForceHTTPS) && present &&
```
ivard, line 257: `ForceHTTPS` is true by default except for in tests. This means that `http://` requestors in the requestor schemes are never replaced with the name, also not when developer mode is enabled in the IRMA app. Was this the intension?

irmaclient/logs.go
davidv1992: Found a better aproach that uses shadowing

irmaclient/session.go
davidv1992: Yes, as SSL provides the needed assurance on identity to make it reasonable to use a pretty verifier name.

schemes.go
davidv1992: Fixed

PR #110, opened 2020-10-02
Impelemented adding of suggestions for non-singletons.

davidv1992:


irmaclient/client.go
```diff
@@ -589,14 +589,24 @@ func (client *Client) credCandidates(request irma.SessionRequest, con irma.Attri
 			// then the entire conjunction is unsatisfiable
 			satisfiable = false
 		}
-		if len(c) == 0 {
-			if client.addCredSuggestion(request, credTypeID) {
-				// No acceptable credentials found. Excluding some nonsensical cases,
-				// add an "empty" credential (i.e. without hash) as a suggestion to the user
-				c = append(c, &credCandidate{Type: credTypeID})
+		// Determine whether we force an attribute value for any attribute requested from this credential here.
```
sietseringers, line 592: To me the "we" makes this comment confusing: it is not "us" but the requestor that requests specific attribute values. I had to read the code below in order to be able to understand the comment, instead of vice versa. Suggestion: `Determine if the session request forces (...)`

irmaclient/client.go
```diff
@@ -589,14 +589,24 @@ func (client *Client) credCandidates(request irma.SessionRequest, con irma.Attri
...
+		haveFixedAttrValue := false
+		for _, attr := range con {
+			if attr.Type.CredentialTypeIdentifier() != credTypeID {
+				continue
+			}
+			if attr.Value != nil {
+				haveFixedAttrValue = true
```
sietseringers, line 599: The `have` here refers to something in the session request, while below the `have` in `haveCandidates` refers to something in our attribute wallet. Upon first glance, this variable name suggests to me that (when true) we have the required value in our wallet, which is not what this means. Suggestion: just `fixedAttrValue`.

sietseringers (changes requested):
I find a comment and variable name a little too unclear, otherwise this looks fine.

irmaclient/client.go
davidv1992: fixed it, good suggestion

irmaclient/client.go
davidv1992: fixed it, good suggestion

PR #115, opened 2020-12-11
Suppress keyshare test logging output

sietseringers:


davidv1992:
I agree that this will prevent the logger from being overwritten, but also reduces test coverage by not actively testing whether the logger generation code gives problems.

davidv1992:
I think the only real long term solution is to actually implement a more sensible approach to logging in the main package other than relying on a global variable, but that will take a lot more effort

davidv1992:
I specifically mean the code in for example app/conf.go lines 111 to 116, which ensures that the logger is set, and that this is done before anything that might call logging code. By never calling this during tests without the logger already set in the configuration, you will miss bugs in this code and in the order of this code with respect to actual calls to the logger.

davidv1992:
I don't agree with you that testing this is not important enough to deal with the additional output, but that is your call to make.

In any case, this could also be fixed in a much better way by reworking the logging solution to not use global variables, at which point the larger output does no longer leak to other tests fixing the issue in all tests where that output isn't relevant.

PR #117, opened 2021-01-11
Add log field 'url' for 404 responses.

synaptic-cleft:


server/api.go
```diff
@@ -414,6 +414,9 @@ func LogResponse(status int, duration time.Duration, binary bool, response []byt
 	l := Logger.WithFields(fields)
 	if status < 400 {
 		l.Trace("<= response")
+	} else if status == 404 {
+		l.WithField("url", url)
+		l.Warn("<= response")
```
sietseringers, line 419: This does not work: `l.WithField()` returns a new logger that you do not use in the next line, so the `url` field does not appear in the emitted log output:

```
[2021-01-11T15:40:04+01:00]  WARN <= response duration=65.231µs status=404
```

Instead, you should say `l.WithField("url", url).Warn("<= response")`.

By the way, you can test this change by starting a local IRMA server using `go run ./irma server`, and then doing some HTTP request (i.e., to `http://$yourIP:8080/nonexistingURL`).

sietseringers (changes requested):

server/api.go
synaptic-cleft: Mh, so true... Will fix and amend the commit.

Thanks for the quick reply :)

sietseringers: Another thing (sorry for not including this earlier), in case of other errors is probably also a good idea. So maybe not branch on 404, but always include the URL in the `else` that was already there.

synaptic-cleft: No worries. Updated.

sietseringers (approved):

PR #119, opened 2021-01-21
Fix exp date for default disclosure request

### LINKED ISSUE #86: resultJwt has exp = iat (immediately expired) if no explicit validity is passed in disclosure request
When I create a disclosure session jwt request and send it to the IRMA server using irmajs, after the disclosure happens and the resultJwt is retrieved, the `exp` field in the jwt is equal to the `iat` field, making the jwt expired as soon as it is issued.

If I add `validity` to the `sprequest` it adds it to the `iat` and it works as intended.

## Example

Request JWT:
```jwt
eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzcHJlcXVlc3QiOnsicmVxdWVzdCI6eyJAY29udGV4dCI6Imh0dHBzOi8vaXJtYS5hcHAvbGQvcmVxdWVzdC9kaXNjbG9zdXJlL3YyIiwiZGlzY2xvc2UiOltbWyJpcm1hLWRlbW8uTWlqbk92ZXJoZWlkLmFnZUxvd2VyLm92ZXIxOCJdXV19fSwiaWF0IjoxNTgyMjc5ODE3LCJpc3MiOiJzc2ktc2VydmljZS1wcm92aWRlciIsInN1YiI6InZlcmlmaWNhdGlvbl9yZXF1ZXN0In0.K_tW_BWzTdVLg4hebmua8GFX64Qwbt_YrYif-rrZLUa3sgVbiuagEhjkqJLWNOm6wVfd056zi6dXm9WRcc6UrL3RgVizJFayVINB6o8dft_CFr7e1ffSU8W4PcukCSiVddWqtKdX8P1OqcMq2e9atyUt2KeWwndMM65eXxtST4ommGZBaiHGcIPM5uFZf9fGFnmxK11gKXE8mKoi0bX8UN-jHhG0pI6qYkATrzv1YwhOLwqz9hcMVMOnxLiQB0u1CJJ6g4nmfwBvTg-Sl59w3TJWUtz48nTJpnGyzeUdDu819hayydPDD0YV2brjmtNtWmJCQNvfwvwLiBUn2WeU2y8EuQknfif0Ol8TDu05QUlf1edULdJr3m6L-MRqIHN24CJaUSevVH50nFAp-FQUISr1GDzbl8uoUw2-s0vg4bdG8YFAJnDvKLs3R687skmA-t2C6QABQDdYxEta8_9TPuRrs9YFrQbX-ddg4bsQPW3vHmxHrM4JMKswPHc9cfHqoOw4llT2PXBkuusDQf5Cg7tKk084GKalX1fTVEb3WlWs9C1l_HVh7bx66nC23kVlDelwy8hyiIw88VM8oBFefm_LEod6DV3I8ZMjgCZh0Dzhhp0MxdntoESvOWMoHlc0_a-i7ixpurWxqiZv1k7jorkFsgBxS2ah9XxUX6Unvcc
```

Request JWT payload:
```json
{
  "sprequest": {
    "request": {
      "@context": "https://irma.app/ld/request/disclosure/v2",
      "disclose": [
        [
          [
            "irma-demo.MijnOverheid.ageLower.over18"
          ]
        ]
      ]
    }
  },
  "iat": 1582279817,
  "iss": "ssi-service-provider",
  "sub": "verification_request"
}
```

Result JWT:
```jwt
eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1ODIyNzg1NzAsImlhdCI6MTU4MjI3ODU3MCwiaXNzIjoiaXJtYXNlcnZlciIsInN1YiI6ImRpc2Nsb3NpbmdfcmVzdWx0IiwidG9rZW4iOiJpVEFOQlhjcUxlZVNnV0xiRlQ2YSIsInN0YXR1cyI6IkRPTkUiLCJ0eXBlIjoiZGlzY2xvc2luZyIsInByb29mU3RhdHVzIjoiVkFMSUQiLCJkaXNjbG9zZWQiOltbeyJyYXd2YWx1ZSI6InllcyIsInZhbHVlIjp7IiI6InllcyIsImVuIjoieWVzIiwibmwiOiJ5ZXMifSwiaWQiOiJpcm1hLWRlbW8uTWlqbk92ZXJoZWlkLmFnZUxvd2VyLm92ZXIxOCIsInN0YXR1cyI6IlBSRVNFTlQiLCJpc3N1YW5jZXRpbWUiOjE1ODA5NDcyMDB9XV19.fAAErAwKkrXD8B9lLv-00FlG_VFAtSp5RgF3KYAo9b9hlWp7O3dmgK2eYTcoFs5EZa4ib6gfD4ADmIdrPQJxkE_vr2PwqLyGPfePc6oiOPFft4hhiMLlldbZV4q-btksHZadznCS3n2ibP2kqT8bR3ehbwXXOmCtc3-8_SDzb1LHpSr-9Ns44OBF7NDnfCMlvkYXckF28QtxYONwQaaGU_Ge795s0OuepgJgGljVtiKtWGe6odrA0Y1FuTzB_xn21IXznembBn-BlNhhRKaiWDjDbDTHibz5Euorz9m9Naqik55lz-Wb5b7sgwN1abNFzs9zzYrGEO-iePh1tDmyH8ORO8n9snZZZSclkGcHij6F8eChzmZEtj4-a9Onxz_FalY834rY0T5_m6EGypyrfLnha9zcpbmuib_tCTNyDWs4ztXlt4MpsBcD8ED0ERsd6OYzo7_TdK-MAhFW6wyDHd6-Xk90wG-xKmCB0gM4qOwIMncX0Zpra1NJFIHzQx8PAYLTR3FWTUvNQhMAM5Fsgjf88vi9kbJakz1mpArX8oqbDugiBv_00M1KCd5E4JPM4JjRnYwc4T7o75En7e_5Q4MdC9zHpodDCBskHMlqV3X3J32amA9oCKtVC0G_M93VqhLuJzX_y4g2qjg0YelySic4XCNmnuz4jJpih232R4E
```

Result jwt payload:
```json
{
  "exp": 1582278570,
  "iat": 1582278570,
  "iss": "irmaserver",
  "sub": "disclosing_result",
  "token": "iTANBXcqLeeSgWLbFT6a",
  "status": "DONE",
  "type": "disclosing",
  "proofStatus": "VALID",
  "disclosed": [
    [
      {
        "rawvalue": "yes",
        "value": {
          "": "yes",
          "en": "yes",
          "nl": "yes"
        },
        "id": "irma-demo.MijnOverheid.ageLower.over18",
        "status": "PRESENT",
        "issuancetime": 1580947200
      }
    ]
  ]
}
```

## Environment

- Latest debian docker image
- 
   ```bash
   $ irma version
   IRMA toolkit v0.4.1
   Documentation: https://irma.app/docs
   
   Version:  0.4.1
   OS/Arg:   linux/amd64
   ``` 
- production mode
- jwt priv key set
- requestors set with public key authmethod

synaptic-cleft:
Fixes #86

### DIFF
--- a/internal/sessiontest/session_test.go
+++ b/internal/sessiontest/session_test.go
@@ -31,13 +31,18 @@ func TestDisclosureSession(t *testing.T) {
 	request := getDisclosureRequest(id)
 	responseString := sessionHelper(t, request, "verification", nil)
 
-	// Check default JWT validity
-	claims := jwt.MapClaims{}
-	jwt.ParseWithClaims(responseString, claims, nil)
+	// Validate JWT
+	claims := struct {
+		jwt.StandardClaims
+		*server.SessionResult
+	}{}
+	_, err := jwt.ParseWithClaims(responseString, &claims, func(token *jwt.Token) (interface{}, error) {
+		return &JwtServerConfiguration.JwtRSAPrivateKey.PublicKey, nil
+	})
+	require.NoError(t, err)
 
-	iat := claims["iat"].(float64)
-	exp := claims["exp"].(float64)
-	require.True(t, iat+120000 == exp)
+	// Check default expiration time
+	require.True(t, claims.IssuedAt+irma.DefaultJwtValidity == claims.ExpiresAt)
 }
 
 func TestNoAttributeDisclosureSession(t *testing.T) {
--- a/internal/sessiontest/session_test.go
+++ b/internal/sessiontest/session_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"github.com/dgrijalva/jwt-go"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -28,7 +29,15 @@ func TestSigningSession(t *testing.T) {
 func TestDisclosureSession(t *testing.T) {
 	id := irma.NewAttributeTypeIdentifier("irma-demo.RU.studentCard.studentID")
 	request := getDisclosureRequest(id)
-	sessionHelper(t, request, "verification", nil)
+	responseString := sessionHelper(t, request, "verification", nil)
+
+	// Check default JWT validity
+	claims := jwt.MapClaims{}
+	jwt.ParseWithClaims(responseString, claims, nil)
+
+	iat := claims["iat"].(float64)
+	exp := claims["exp"].(float64)
+	require.True(t, iat+120000 == exp)
 }
 
 func TestNoAttributeDisclosureSession(t *testing.T) {
--- a/server/api.go
+++ b/server/api.go
@@ -189,13 +189,23 @@ func WriteString(w http.ResponseWriter, str string) {
 //  - SessionRequest instances (DisclosureRequest, SignatureRequest, IssuanceRequest)
 //  - JSON representations ([]byte or string) of any of the above.
 func ParseSessionRequest(request interface{}) (irma.RequestorRequest, error) {
+	rr, e := ParseInput(request)
+	if e != nil {
+		return nil, e
+	}
+	rr.Base().SetDefaultsIfNecessary()
+
+	return rr, e
+}
+
+func ParseInput(request interface{}) (irma.RequestorRequest, error) {
 	switch r := request.(type) {
 	case irma.RequestorRequest:
 		return r, nil
 	case irma.SessionRequest:
 		return wrapSessionRequest(r)
 	case string:
-		return ParseSessionRequest([]byte(r))
+		return ParseInput([]byte(r))
 	case []byte:
 		var attempts = []irma.Validator{&irma.ServiceProviderRequest{}, &irma.SignatureRequestorRequest{}, &irma.IdentityProviderRequest{}}
 		t, err := tryUnmarshalJson(r, attempts)
--- a/server/api.go
+++ b/server/api.go
@@ -189,7 +189,7 @@ func WriteString(w http.ResponseWriter, str string) {
 //  - SessionRequest instances (DisclosureRequest, SignatureRequest, IssuanceRequest)
 //  - JSON representations ([]byte or string) of any of the above.
 func ParseSessionRequest(request interface{}) (irma.RequestorRequest, error) {
-	rr, e := ParseInput(request)
+	rr, e := parseInput(request)
 	if e != nil {
 		return nil, e
 	}
@@ -198,14 +198,14 @@ func ParseSessionRequest(request interface{}) (irma.RequestorRequest, error) {
 	return rr, e
 }
 
-func ParseInput(request interface{}) (irma.RequestorRequest, error) {
+func parseInput(request interface{}) (irma.RequestorRequest, error) {
 	switch r := request.(type) {
 	case irma.RequestorRequest:
 		return r, nil
 	case irma.SessionRequest:
 		return wrapSessionRequest(r)
 	case string:
-		return ParseInput([]byte(r))
+		return parseInput([]byte(r))
 	case []byte:
 		var attempts = []irma.Validator{&irma.ServiceProviderRequest{}, &irma.SignatureRequestorRequest{}, &irma.IdentityProviderRequest{}}
 		t, err := tryUnmarshalJson(r, attempts)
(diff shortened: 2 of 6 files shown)


server/api.go
```diff
@@ -189,13 +189,23 @@ func WriteString(w http.ResponseWriter, str string) {
...
+	rr, e := ParseInput(request)
+	if e != nil {
+		return nil, e
+	}
+	rr.Base().SetDefaultsIfNecessary()
+
+	return rr, e
+}
+
+func ParseInput(request interface{}) (irma.RequestorRequest, error) {
```
sietseringers, line 201: Since this function is only used here and additionally not necessarily suitable for use by others, I recommend to not export it (i.e. rename to `parseInput`).

internal/sessiontest/session_test.go
```diff
@@ -28,7 +29,15 @@ func TestSigningSession(t *testing.T) {
 func TestDisclosureSession(t *testing.T) {
 	id := irma.NewAttributeTypeIdentifier("irma-demo.RU.studentCard.studentID")
 	request := getDisclosureRequest(id)
-	sessionHelper(t, request, "verification", nil)
+	responseString := sessionHelper(t, request, "verification", nil)
+
+	// Check default JWT validity
+	claims := jwt.MapClaims{}
+	jwt.ParseWithClaims(responseString, claims, nil)
```
sietseringers, line 36: This ignores the error parameter that this function can return; you should catch it and feed it to `require.NoError`.

internal/sessiontest/session_test.go
```diff
@@ -28,7 +29,15 @@ func TestSigningSession(t *testing.T) {
...
+
+	iat := claims["iat"].(float64)
+	exp := claims["exp"].(float64)
+	require.True(t, iat+120000 == exp)
```
sietseringers, line 40: - If the JWT would put something else than floats in here, then these type assertions would panic. You can write `exp, ok := claims["exp"].(float64)` and then check `ok` first to check that we indeed received a float (and fail the test if not).
- Using equality checks on floats is generally not a good idea because rounding errors may bite you.

In this case, we can avoid both issues because we know exactly the structure of the data that should be inside the JWT. The following should work:
```go
claims := struct {
	jwt.StandardClaims
	*server.SessionResult
}
_, err = jwt.ParseWithClaims(responseString, claims, nil)
```
Then you can refer to the `iat` and `exp` fields by `claims.IssuedAt` and `claims.ExpiresAt` and receive `int64`'s (and as an extra, you could require the presence of some of the other expected fields).

sietseringers, line 40: Why the factor 1000 (i.e. not 120)? The `iat` and `exp` fields of JWTs are always in seconds according to the spec, and the `ResultJwtValidity` field of the `RequestorBaseRequest` is also in seconds according to its documentation.

server/api_test.go
```diff
@@ -43,6 +43,12 @@ func TestParseSessionRequest(t *testing.T) {
 			res.SessionRequest().Disclosure().Disclose[0][0][0].Type.String())
 	})
 
+	t.Run("correct default validity", func(t *testing.T) {
+		res, err := ParseSessionRequest(requestJson)
+		require.NoError(t, err)
+		require.Equal(t, 120, res.Base().ResultJwtValidity)
```
sietseringers, line 49: Here and elsewhere across this PR it would be nice if the magic number `120` would be replaced by the new constant `irma.DefaultJwtValidity`.

server/api.go
synaptic-cleft: done

internal/sessiontest/session_test.go
synaptic-cleft: Catching the error is nicer indeed. It was ignored as I first only checked for the difference in expiration time and issueing time. In a new commit, a function with the corresponding public key is passed to the ParseWithClaims and the error is taken care of.

internal/sessiontest/session_test.go
synaptic-cleft: Pushed new commit with custom claims resulting in ints as you suggested.

synaptic-cleft: Totally stupid brainfart. Corrected to original expiration time calculation.

server/api_test.go
synaptic-cleft: I changed it here now but couldn't find more occurrences that I hadn't yet replaced earlier.

PR #120, opened 2021-01-25
Verify requestor scheme consistency during scheme parsing

sietseringers:


schemes.go
```diff
@@ -921,6 +925,71 @@ func (scheme *SchemeManager) parseContents(conf *Configuration) error {
...
+	allCredTypesForScheme := map[CredentialTypeIdentifier]CredentialType{}
+	for _, credType := range conf.CredentialTypes {
+		if credType.SchemeManagerID == scheme.ID {
+			allCredTypesForScheme[credType.Identifier()] = *credType
+		}
+	}
+
+	// validate that there are no circular dependencies
+	for _, item := range allCredTypesForScheme {
+		if err := item.validateDependencies(conf, []CredentialTypeIdentifier{}); err != nil {
```
sietseringers, line 943: Instead of having a second loop here, would it not be possible to invoke `validateDependencies()` in the first loop above within the if?

schemes.go
```diff
@@ -921,6 +925,71 @@ func (scheme *SchemeManager) parseContents(conf *Configuration) error {
...
+						" has dependency outside the scheme: " + conf.CredentialTypes[item].Identifier().String())
+				}
+
+				if ok := validatedDeps.contains(item); ok {
+					return errors.New("circular dependency " + item.String() + " found in: " + validatedDeps.String())
+				}
+
+				if conf.CredentialTypes[item].Dependencies != nil {
+					validatedDeps = append(validatedDeps, item)
+					return conf.CredentialTypes[item].validateDependencies(conf, validatedDeps)
```
sietseringers, line 969: This will make the function exit as soon as it encounters a credential type that itself has dependencies, even if there are more credentials present in `conf.CredentialTypes[ct.Identifier()].Dependencies`. Did you mean to return only in case of error, instead of always?

Additionally, this seems to ignore the disjunction structure: it enforces that no credential type occurs a second time anywhere, if it is found once anywhere within the `Dependencies`. But if credential type X occurs in disjunction 1, then it would be fine if it also occurs in disjunction 2, or in dependencies of credential types of disjunction 2.

descriptions.go
```diff
@@ -356,10 +357,42 @@ func (wizard *IssueWizard) Validate(conf *Configuration) error {
...
 	for i, outer := range wizard.Contents {
 		for j, middle := range outer {
 			for k, item := range middle {
-				if err := item.Validate(conf); err != nil {
+				// validate all non-credential-items of a wizard are at the end
+				if item.Type != "credential" {
+					shouldBeLast = true
+				} else {
+					if shouldBeLast {
+						return errors.Errorf("non-credential types in wizard " + wizard.ID + " should come last")
```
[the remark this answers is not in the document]
synaptic-cleft: Good point, done.

descriptions.go
```diff
@@ -374,7 +407,90 @@ func (wizard *IssueWizard) Validate(conf *Configuration) error {
...
+	}
+
+	all = append(all, choice)
+
+	return all
+}
+
+func userHasCreds(items []IssueWizardItem, creds map[CredentialTypeIdentifier]struct{}) bool {
+	for _, val := range items {
+		if _, ok := creds[*val.Credential]; !ok {
```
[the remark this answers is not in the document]
synaptic-cleft: packed it in another if instead of a panic. good enough?

descriptions.go
```diff
@@ -522,6 +696,15 @@ func (ts *TranslatedString) UnmarshalXML(d *xml.Decoder, start xml.StartElement)
 	return nil
 }
 
+func (ts *TranslatedString) validate(errorMsg string) error {
```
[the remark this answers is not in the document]
synaptic-cleft: I abstracted the method so it can be used by validateTranslations() and put a comment there that it therefore cannot be empty anymore. For what I can tell, it's logic to receive a warning when the value is empty there.

schemes.go
synaptic-cleft: yeah, that part was utterly useless. fixed now

irmago_test.go
```diff
@@ -985,6 +985,371 @@ func TestWizardFromScheme(t *testing.T) {
...
+			{{{
+				Type:       IssueWizardItemTypeCredential,
+				Credential: credidptr("scheme.issuer.e"),
+				Text:       &TranslatedString{"en": "custom description of credential e"},
+			}}},
+		},
+	}
+
+	err := wizard.Validate(conf)
+	require.Equal(t, "non-credential types in wizard testwizard should come last", err.Error())
```
sietseringers, line 1073: For requiring errors with specific error messages you can use `require.EqualError()` (below too).

irmago_test.go
```diff
@@ -985,6 +985,371 @@ func TestWizardFromScheme(t *testing.T) {
...
+			credid("scheme.issuer.c"): credtype("scheme.issuer.c",
+				"scheme.issuer.a", "scheme.issuer.b",
+			),
+			credid("scheme.issuer.d"): credtype("scheme.issuer.d",
+				"scheme.issuer.a",
+			),
+			credid("scheme.issuer.e"): credtype("scheme.issuer.e",
+				"scheme.issuer.d",
+			),
+			credid("scheme.issuer.e"): credtype("scheme.issuer.e"),
```
sietseringers, line 1092: This overwrites the insertion of `scheme.issuer.e` four lines above.

irmago_test.go
```diff
@@ -985,6 +985,371 @@ func TestWizardFromScheme(t *testing.T) {
...
+	conf := &Configuration{CredentialTypes: map[CredentialTypeIdentifier]*CredentialType{}}
+	for _, discon := range dependencies {
+		for _, con := range discon {
+			for _, cred := range con {
+				conf.CredentialTypes[cred] = credtype(cred.String())
+			}
+		}
+	}
+
+	credTypId := NewCredentialTypeIdentifier("a.a.a")
```
sietseringers, line 1172: For consistency it would be better to use `credid()` here.

irmago_test.go
```diff
@@ -985,6 +985,371 @@ func TestWizardFromScheme(t *testing.T) {
...
+			}}},
+		},
+	}
+
+	err := wizard.Validate(conf)
+	require.Equal(t, "wizard with wizard ID testwizard too complex", err.Error())
+}
+
+func TestIssueWizardItemValidation(t *testing.T) {
+	dependencies := CredentialDependencies{
```
sietseringers, line 1152: Apart from populating `Configuration.CredentialTypes` below, you don't seem to use `dependencies at all`. Would it not be better to remove this and populate with just the two credential types which are used in the test? Same goes for `TestIssueWizardFAQSummariesValidation` below.

schemes.go
synaptic-cleft: Right, the check was not implemented properly. As discussed orally, the dependency check should pass if there is at least one valid dependency flow.

irmago_test.go
```diff
@@ -985,6 +985,375 @@ func TestWizardFromScheme(t *testing.T) {
...
+	conf.CredentialTypes[credTypId].Dependencies = CredentialDependencies{
+		{
+			{credid("a.b.a")},
+		},
+	}
+	require.Equal(t, "FAQSummary missing for last item in chain: a.a.a, a.b.a", tester.validate(conf).Error())
+}
+
+func TestCircularDependenciesValidation(t *testing.T) {
+	credTypeIdA := credid("scheme.issuer.a")
```
sietseringers, line 1235: When putting acronyms in variable names there is a mild preference for using all caps for the acronym, see e.g. https://github.com/golang/go/wiki/CodeReviewComments#initialisms. (This is, however, not consistently done across irmago).

schemes.go
```diff
@@ -921,6 +925,84 @@ func (scheme *SchemeManager) parseContents(conf *Configuration) error {
...
+				// all items need to be valid for middle to be valid
+				if toBeChecked == item {
+					innerConSatisfied = false
+					break
+				}
+
+				if conf.CredentialTypes[item].Dependencies != nil {
+					if e := conf.CredentialTypes[item].validateDependencies(conf, append(validatedDeps, item), toBeChecked); e != nil {
+						if e.Error() == "No valid dependency branch could be built. There might be a circular dependency." {
+							innerConSatisfied = false
```
sietseringers, line 970: In this case you can `break` like in line 964, right?

schemes.go
```diff
@@ -921,6 +925,84 @@ func (scheme *SchemeManager) parseContents(conf *Configuration) error {
...
+						} else {
+							return e
+						}
+					}
+				}
+			}
+
+			if innerConSatisfied {
+				atLeastOneInnerConSatisfied = true
```
sietseringers, line 979: Here too you can `break` now, right?

schemes.go
```diff
@@ -921,6 +925,84 @@ func (scheme *SchemeManager) parseContents(conf *Configuration) error {
...
+	return nil
+}
+
+func (ct CredentialType) validateDependencies(conf *Configuration, validatedDeps DependencyChain, toBeChecked CredentialTypeIdentifier) error {
+	if len(validatedDeps) >= maxDepComplexity {
+		return errors.New("dependency tree too complex: " + validatedDeps.String())
+	}
+	for _, outer := range conf.CredentialTypes[ct.Identifier()].Dependencies {
+		// at least one middle needs to be valid
+		atLeastOneInnerConSatisfied := false
```
sietseringers, line 951: You might consider to use the same variable names for these booleans as `IssueWizardContents.Choose()`, which uses two booleans in exactly the same fashion - or rename those to these.

schemes.go
synaptic-cleft: introducing a break here would not further have the rest of the branches not be validated for being too complex

PR #122, opened 2021-02-05
Force Unix (LF) line endings in schemes during irma scheme sign

sietseringers:
If a text file in a scheme has CRLF line endings, the following error is now emitted:
```
[2021-02-05T15:55:18+01:00] ERROR Failed to sign scheme: Failed to calculate file index: irma-demo/description.xml contains CRLF (Windows) line endings, please convert to LF
exit status 1
```


ivard (approved):
Looks good to me. Maybe one small question, is it really necessary that `.sig` files are excluded? They are never included in the index file anyway right?

PR #124, opened 2021-02-08
Add flag to IRMA server allowing callbackUrl in session requests without JWT private key installed

### LINKED ISSUE #123: sessions with callbackUrl are incorrectly rejected when no JWT is installed
According to the documentation, specifically (https://irma.app/docs/irma-server/#static-irma-qrs), callbackUrls can be used even when no JWT key is installed, however on irma server version 0.6.1 such requests are rejected anyway.

I would argue this behaviour is also unwanted since there are multiple ways such a setup can be secure even if there is no direct authentication of the irma server on the callback url, most notably urls only reachable from the server and by only using the token from the request and refetching the result from the irma server upon receiving the callback.

sietseringers:
Fixes #123.


ivard (approved):
Looks good to me.

PR #125, opened 2021-02-08
In static sessions, also require allow_unsigned_requests to be set if no JWT private key installed

sietseringers:


ivard (approved):
Yeah that seems reasonable indeed. Otherwise it would be a bit inconsistent.

PR #127, opened 2021-02-09
Ensure callback body content type is json for json messages.

davidv1992:
Fixes a bug that caused the callback to indicate content type of the request body as text even when no jwt key was installed in the server.

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -18,7 +18,7 @@ import (
 	"github.com/dgrijalva/jwt-go"
 	"github.com/go-chi/chi/middleware"
 	"github.com/go-errors/errors"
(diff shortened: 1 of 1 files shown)


sietseringers (changes requested):
Thanks for noticing and fixing this.

Your change fixes the problem by adding code duplication. However, is this not fixable by removing code instead, as follows?
* change the type of `res` to `interface{}`
* in the `else` branch, remove the JSON serialization and just say `res = result`.

davidv1992:
Fair point, will make the change

davidv1992:
@sietseringers fixed

PR #129, opened 2021-02-11
Fix panic in scheme downloading on Windows due to usage wrong path separator

sietseringers:


ivard (approved):
Looks good to me.

Are we sure this problem does not occur on more places?

sietseringers:
I thought I had this fixed at all places (but this bug proves that I was wrong). Merging this now to solve it here; if it pops up at other places we'll fix it then.

Merged in 8a11460a281c62e811c05fcb6976ca513ed2dfbf.

PR #131, opened 2021-02-12
Create separate LogoPath field in RequestorInfo for file path

ivard:
For consistency, we also store the LogoPath in a separate field for `RequestorInfo`, just like we did for wizards (PR #118).

### DIFF
--- a/schemes.go
+++ b/schemes.go
@@ -1193,6 +1193,13 @@ func (scheme *RequestorScheme) setPath(path string) { scheme.storagepath = path 
 
 func (scheme *RequestorScheme) parseContents(conf *Configuration) error {
 	for _, requestor := range scheme.requestors {
+		if requestor.Logo != nil {
+			logoPath := filepath.Join(scheme.path(), "assets", *requestor.Logo+".png")
+			if exists, _ := common.PathExists(logoPath); !exists {
+				return errors.Errorf("Asset could not be found for logo %s", requestor.Logo)
+			}
+			requestor.LogoPath = &logoPath
+		}
 		for _, hostname := range requestor.Hostnames {
 			if _, ok := conf.Requestors[hostname]; ok {
 				return errors.Errorf("Double occurence of hostname %s", hostname)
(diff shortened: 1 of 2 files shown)


schemes.go
```diff
@@ -1193,6 +1193,13 @@ func (scheme *RequestorScheme) setPath(path string) { scheme.storagepath = path
 
 func (scheme *RequestorScheme) parseContents(conf *Configuration) error {
 	for _, requestor := range scheme.requestors {
+		if requestor.Logo != nil {
+			logoPath := filepath.Join(scheme.path(), "assets", *requestor.Logo+".png")
+			if exists, _ := common.PathExists(logoPath); !exists {
+				return errors.Errorf("Asset could not be found for logo %s", requestor.Logo)
+			}
+			requestor.LogoPath = &logoPath
```
sietseringers, line 1201: This does not validate that the logo is correctly signed, by checking that the contents of the logo hashes to the hash in the filename.

(When introducing requestor schemes one of the design goals was to make `RequestorInfo` instances portable; i.e. make it such that in the future we might sign an individual `RequestorInfo` and give that to the requestor, who can then include it in the session request. For that reason, the logo is not in the scheme index so it contents must be manually validated when read.)

In the `wizard` branch I created a method `checkLogo()` on `RequestorScheme` which does that, as well as what you do here. Perhaps you could copy and use that method?

ivard: Ah I kind of assumed that this was already checked and did not think about this. But indeed that is something that should be added.

ivard: I just checked, and the check is already there. The check is done on [line 1252 in the `validate` function](https://github.com/privacybydesign/irmago/blob/e5002f5f84bd2fbbc31c8d2af93ca9adfd4210f5/schemes.go#L1252). `validate` is called just before `parseContents`, so the hash is always checked.

There only was a little bit of code duplication for composing the logo file path. I removed that by combining this in one function. I thought it was nice to keep clean the `validate` function, such that it only does checking. The actual assignment to `LogoPath` is only done in `parseContents` to make sure that `validate` does not make modifications.

sietseringers (approved):

PR #138, opened 2021-03-04
Fix scheme restoration to incorrect path in case of error 

sietseringers:


ivard (approved):
Subtile error, but this seems to fix it indeed.

PR #141, opened 2021-03-16
Simplify revocation commands of irma CLI

sietseringers:
The revocation subcommands are now subcommands of irma issuer.


ivard (approved):
Looks good to me.

PR #142, opened 2021-03-17
Respond with a 204 HTTP status code after cancelling a Session

### LINKED ISSUE #140: Cancel a IRMA session
Whenever you cancel an IRMA session (by sending DELETE `/session/[:token]` to the IRMA server), you get a `200` response with an empty from the server. 

A common design pattern for DELETE response is to either (1) use the `204` HTTP status code with an empty body or (2) a `200` HTTP status code with the deleted session status as response. The current implementation seems to mix both design patterns, which leads to (IMO) an unexpected response for cancelling an IRMA session.

> If a DELETE method is successfully applied, the origin server SHOULD send a 202 (Accepted) status code if the action will likely succeed but has not yet been enacted, a 204 (No Content) status code if the action has been enacted and no further information is to be supplied, or a 200 (OK) status code if the action has been enacted and the response message includes a representation describing the status. - [The original HTTP/1.1 spec](https://tools.ietf.org/html/rfc7231)

I was wondering; is this design on purpose, or might it be open to be improved?

stefanvermaas:
According to the official HTTP/1.1 specs a successful DELETE request can be handled in various ways.

> If a DELETE method is successfully applied, the origin server SHOULD send a 202 (Accepted) status code if the action will likely succeed but has not yet been enacted, a 204 (No Content) status code if the action has been enacted and no further information is to be supplied, or a 200 (OK) status code if the action has been enacted and the response message includes a representation describing the status.

This will make sure a 204 HTTP status code is returned after successfully cancelling the session and it fixes #140 

I assumed the session will be cancelled right-away and it's not some sort of asynchronous task. If that do is the case (the session is not yet cancelled when the response is already returned) we should change the HTTP response code to a `202`.

I'm looking for some help on writing the tests for the HTTP status code. It seems there are none yet, but I'm not entirely sure what is the best way to add them and how to add them.

### DIFF
--- a/server/irmaserver/handle.go
+++ b/server/irmaserver/handle.go
@@ -12,7 +12,7 @@ import (
 	"github.com/go-chi/chi"
 	"github.com/privacybydesign/gabi"
 	"github.com/privacybydesign/gabi/signed"
-	"github.com/privacybydesign/irmago"
+	irma "github.com/privacybydesign/irmago"
 	"github.com/privacybydesign/irmago/internal/common"
 	"github.com/privacybydesign/irmago/server"
 	"github.com/sirupsen/logrus"
@@ -284,7 +284,7 @@ func (s *Server) handleSessionStatusEvents(w http.ResponseWriter, r *http.Reques
 
 func (s *Server) handleSessionDelete(w http.ResponseWriter, r *http.Request) {
 	r.Context().Value("session").(*session).handleDelete()
-	w.WriteHeader(200)
+	w.WriteHeader(204)
 }
 
 func (s *Server) handleSessionGet(w http.ResponseWriter, r *http.Request) {


server/irmaserver/handle.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/go-chi/chi"
 	"github.com/privacybydesign/gabi"
 	"github.com/privacybydesign/gabi/signed"
-	"github.com/privacybydesign/irmago"
+	irma "github.com/privacybydesign/irmago"
```
stefanvermaas, line 15: Hmm it seems this is added by `go` itself. Should I remove this?

ivard:
This is indeed a good idea. More of a technical detail before this can be merged. The IRMA client (the app) also uses this endpoint and that part of the code cannot deal with 204's at the moment. I introduce support for this in #98, so after that one is merged, we can consider merging this too.

server/irmaserver/handle.go
sietseringers: No, it's fine; we have these changes also in other PRs/commits. Apparently they changed `gofmt` at some point to do this.

stefanvermaas:
> I introduce support for this in #98, so after that one is merged, we can consider merging this too.

Sounds good! Let's wait for it to finish. In the meantime it would make sense to add tests, but I'm not entirely sure how to add these. Do you have a good resource for writing tests for go applications or would one of you be willing to add tests to this PR?

ivard:
We already have some [unit tests in irmago](https://github.com/privacybydesign/irmago#running-the-unit-tests). The testing code is situated [here](https://github.com/privacybydesign/irmago/tree/master/internal/sessiontest).

It would maybe be nice to have a unit test that starts a session, cancels it at some point and checks whether it is properly deleted. I cannot find such an unit test now indeed. That test should be in [this set of tests](https://github.com/privacybydesign/irmago/blob/master/internal/sessiontest/session_test.go) I think.

stefanvermaas:
Would you mind explaining how I can create a IRMA disclosure session? I've been messing around with the tests and tried to find references on how to do it, but I can't get it to work :)!  Any help would be appreciated.

ivard:
Ah now you send the `DELETE` request manually. It also an option to use the `irmaclient` code (for the IRMA app) for this. Then we test that part of the code too. Of course this test will fail then because, like I said, the client does not support this yet.

You can maybe look to [`sessionHelper`](https://github.com/privacybydesign/irmago/blob/68675054e9871f644ce07f42061fa61fdb5cb98f/internal/sessiontest/main_test.go#L192) function. This helper starts a IRMA server and an irma client (the backend of the app) and performs a session. This one uses a `TestHandler` which stubs all user input of the app for the happy flow. We can make a new handler with the same behaviour that cancels the session.

```
type CancelTestHandler struct {
	TestHandler
}

func (th *CancelTestHandler) Success(result string) {
	th.Failure(&irma.SessionError{ErrorType: irma.ErrorType("Session was not cancelled as expected")})
}

func (th *CancelTestHandler) RequestVerificationPermission(request *irma.DisclosureRequest, satisfiable bool, candidates [][]irmaclient.DisclosureCandidates, ServerName *irma.RequestorInfo, callback irmaclient.PermissionHandler) {
	callback(false, nil)
}
```

Disclaimer: I did not test this code, so there can be mistakes in it :)

PR #144, opened 2021-04-13
Bump gabi to rangeproof supporting version

sietseringers:
Additionally:

* `gabi.PublicKey`, `gabi.PrivateKey` and related methods moved to `gabikeys`
* handle additional errors returned by gabi

### DIFF
--- a/revocation.go
+++ b/revocation.go
@@ -17,8 +17,8 @@ import (
 	"github.com/go-errors/errors"
 	"github.com/hashicorp/go-multierror"
 	"github.com/jinzhu/gorm"
-	"github.com/privacybydesign/gabi"
 	"github.com/privacybydesign/gabi/big"
+	"github.com/privacybydesign/gabi/gabikeys"
 	"github.com/privacybydesign/gabi/revocation"
 	"github.com/privacybydesign/gabi/signed"
 	sseclient "github.com/sietseringers/go-sse"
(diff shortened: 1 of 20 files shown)


revocation.go
```diff
@@ -641,11 +641,7 @@ func (rs *RevocationStorage) SaveIssuanceRecord(id CredentialTypeIdentifier, rec
 	if settings.RevocationServerURL == "" {
 		return errors.New("cannot send issuance record: no server_url configured")
 	}
-	rsk, err := sk.RevocationKey()
-	if err != nil {
-		return err
-	}
-	return rs.client.PostIssuanceRecord(id, rsk, rec, settings.RevocationServerURL)
+	return rs.client.PostIssuanceRecord(id, sk, rec, settings.RevocationServerURL)
```
synaptic-cleft, line 644: Aren't you passing a different private key now? If I follow the refactoring in gabi correctly, the `PrivateKey` contained a field `revocationKey` which was also of type `PrivateKey`. This was retrieved here by `sk.RevocationKey()`.

This `revocationKey` was refactored in gabi and the revocation key is now parsed to the field `ECDSA`. So I would expect that we pass `sk.ECDSA` here.

Am I missing a nesting step that the latest gabi version resolved?

sietseringers: `sk.RevocationKey()` used to return a `revocation.PrivateKey`, which contained part of the data contained in `gabi.PrivateKey`. Now that the latter has moved to `gabikeys.PrivateKey`, it was no longer necessary for `revocation.PrivateKey` to exist, as the `revocation` package can now reference `gabikeys.PrivateKey`, so I refactored away `revocation.PrivateKey` and thus also the `RevocationKey()` method (in a commit which is already merged in `gabi`).

We could indeed pass `sk.ECDSA` to `PostIssuanceRecord` instead of the entire `sk`, but that has little to do with all of the refactoring going on: we could have done that before all refactors as well. I'm not agains it per se, but to me it seems a little better to just keep passing the `sk` and have `PostIssuanceRecord` figure out itself what it needs to do with that (i.e. grab `ECDSA`).

PR #147, opened 2021-05-12
Adds API-prefix 

awesterb:
Adds an `api-prefix` (`-a`) flag to to the irma server, so that `irma server -a /someprefix` will accept `POST /someprefix/session` instead of `POST /session`, etc..

Using this option one can (more easily) run an irma server behind `https://provider.com/irmaserver/` instead of resorting to `https://irmaserver.provider.com`.


server/requestorserver/server.go
```diff
@@ -159,11 +159,22 @@ var corsOptions = cors.Options{
 	AllowedMethods: []string{http.MethodGet, http.MethodPost, http.MethodDelete},
 }
 
+func (s *Server) createRouters() (rootRouter, apiRouter *chi.Mux) {
+	rootRouter = chi.NewRouter()
+	rootRouter.Route(s.conf.ApiPrefix, func(r chi.Router) {
```
sietseringers, line 164: When `s.conf.ApiPrefix` does not start with `/` this panics; this is better checked during validation of the configuration.

server/requestorserver/server.go
```diff
@@ -159,11 +159,22 @@ var corsOptions = cors.Options{
...
+		// r will be a *chi.Mux, see https://github.com/go-chi/chi/blob/c9e87efe9691a63d6a89de8bbd16b04fe4d6640e/mux.go#L262
+		// We don't use a chi.Router, because we need the {NotFound,MethodNotAllowed}Handlers methods of *chi.Mux later on that
+		// are not part of the chi.Router interface.
+		apiRouter = r.(*chi.Mux)
```
sietseringers, line 168: I think we can avoid this type assertion, as wel as simplify the implementation of this feature as follows. Based on the original code:
* Write a new unexported method on `Server` that, given a `router`, either just returns it if `s.conf.ApiPrefix` is empty; or it creates and returns a new `rootRouter` to which the `router` is attached using `rootRouter.Mount(s.conf.ApiPrefix, router.Handler())`;
* Return the return value of that method at the end of `Server.ClientHandler()` and `Server.Handler()`.

sietseringers (changes requested):
Thanks! I think this is a nice feature. I have left some comments in the code.

server/requestorserver/server.go
awesterb: The panic chi causes is already pretty clear:
```
% irma server  -a someapiprefix                         
[...]
panic: chi: routing pattern must begin with '/' in 'someapiprefix/*'
```

server/requestorserver/server.go
awesterb: Thanks, that's much better;  I've followed your suggestion.

server/requestorserver/server.go
sietseringers: True. But:
* Some cleanup code (eg. close database connections) may not get called,
* it breaks a pattern: in lots of other cases of incorrect input the server also does not panic, but instead emits an error message (in `Configuration.initialize()` in `server/requestorserver/conf.go`),
* to me, "panic" makes this look much scarier than it is.

server/requestorserver/server.go
sietseringers: Thanks, this is better. I noticed you mount `router.Handler()` unconditionally to a new router. In most cases `ApiPrefix` will be empty, but this will cause a (small) performance penalty anyway, right? Would it not be better to just return the router in the new method in this case?

server/requestorserver/server.go
awesterb: Ah, right, you rightly prefer an error over a panic.  I've added the check to the configuration initialisation function.

sietseringers (approved):

PR #150, opened 2021-05-20
Support session chaining in requestor api (SessionResult)

leonbotros:
This PR adds changes to the requestor API such that requestors get informed about a possible `nextSession`.
The requestor receives the `requestorToken` of the next session in the `SessionResult`.


server/api.go
```diff
@@ -41,6 +41,7 @@ type SessionResult struct {
 	Disclosed   [][]*irma.DisclosedAttribute `json:"disclosed,omitempty"`
 	Signature   *irma.SignedMessage          `json:"signature,omitempty"`
 	Err         *irma.RemoteError            `json:"error,omitempty"`
+	NextSession irma.RequestorToken          `json:"nextSession,omitempty"`
```
ivard, line 44: Wouldn't you include the whole `SessionPackage` instead of only the requestor token? I can image that for some cases it can also be handy to know the `sessionPtr` too.

ivard (changes requested):
Looks like a neat, small change. I only had one comment.

server/api.go
```diff
@@ -334,6 +335,10 @@ func DoResultCallback(callbackUrl string, result *SessionResult, issuer string,
 
 	var x string // dummy for the server's return value that we don't care about
 	if err := irma.NewHTTPTransport(callbackUrl, false).Post("", &x, res); err != nil {
+		if sessErr, ok := err.(*irma.SessionError); ok && sessErr.RemoteStatus == http.StatusNoContent {
+			// If we get a 204 during this callback that's perfectly fine
+			return
+		}
```
ivard, line 341: Since we ignore the result `x` here, I think it should also work to remove `x` here and say `Post("", nil, res)`. This means we ignore the result. `Post` is implemented in such a way that if a 204 is received and the result is nil, then no error is returned. This means we wouldn't have to manually catch this error here.

server/api.go
ivard: As discussed in Slack, only the `RequestorToken` should work.

server/api.go
leonbotros: You're right. I'll change that.

ivard: Yeah it seems logical behaviour to me to also add 204 support for the normal callback option then indeed. Good catch.

ivard (approved):
I rebased it on master, since it is not really related to `shoulder-surf`. Therefore, I added the changes in `transport.go` to this PR to introduce 204 support.

PR #153, opened 2021-05-31
Support custom text color in issuance wizard headers

sietseringers:


descriptions.go
```diff
@@ -362,6 +363,9 @@ func (wizard *IssueWizard) Validate(conf *Configuration) error {
 	if wizard.Color != nil && !regexp.MustCompile("^#[0-9A-F]{6}$").MatchString(*wizard.Color) {
 		return errors.New("invalid wizard color: must be of the form #RRGGBB")
 	}
+	if wizard.TextColor != nil && !regexp.MustCompile("^#[0-9A-F]{6}$").MatchString(*wizard.TextColor) {
+		return errors.New("invalid wizard text color: must be of the form #RRGGBB")
+	}
```
ivard, line 368: This double negation confuses me a bit. The `^` tag in the regex also specifies a negation right? So this error should be thrown if it does not match the no-match?

ivard (commented):
The change looks good to me. I was only a bit confused by the negation in the regex expression. Is this one fully correct?

descriptions.go
sietseringers: No, `^` in a regexp is not an inversion; instead it matches only the beginning of the string.

ivard: Ah right, you're indeed right. I must have misread the docs there then.

ivard (approved):

descriptions.go
sietseringers: You were perhaps confused with when `^` occurs within square brackets, i.e. a [character class](https://golang.org/pkg/regexp/syntax/#hdr-Syntax). In that case it does act as an inversion.

PR #154, opened 2021-05-31
Prevent possible mutex deadlock when using chained sessions

sietseringers:
When periodically iterating over sessions to see if they need to be expired,
the `memorySessionStore.deleteExpired()` method of the IRMA server locks the
session store mutex, and while that is locked it additionally locks each
session mutex when handling it. If a session is currently being handled by an
endpoint which also locks the session mutex, `deleteExpired()` therefore has to
wait for the handler to finish dealing with the session. However, when that
session attempts to starts a new session (as part of a session chain), it tries
to lock the session store mutex again, which was already locked: a deadlock.

We solve this by making `deleteExired()` not request a session lock while the
entire store is also locked: the store and session locking are now done
sequentially. That way, even when a session is locked by an endpoint in the
loop over the sessions in `deleteExpired()`, that session is free to write new
sessions to the session store.

### DIFF
--- a/server/irmaserver/sessions.go
+++ b/server/irmaserver/sessions.go
@@ -113,10 +113,15 @@ func (s *memorySessionStore) deleteExpired() {
 	// First check which sessions have expired
 	// We don't need a write lock for this yet, so postpone that for actual deleting
 	s.RLock()
-	expired := make([]string, 0, len(s.requestor))
+	toCheck := make(map[string]*session, len(s.requestor))
 	for token, session := range s.requestor {
-		session.Lock()
+		toCheck[token] = session
+	}
+	s.RUnlock()
 
+	expired := make([]string, 0, len(toCheck))
+	for token, session := range toCheck {
+		session.Lock()
 		timeout := maxSessionLifetime
 		if session.status == server.StatusInitialized && session.rrequest.Base().ClientTimeout != 0 {
 			timeout = time.Duration(session.rrequest.Base().ClientTimeout) * time.Second
@@ -134,7 +139,6 @@ func (s *memorySessionStore) deleteExpired() {
 		}
 		session.Unlock()
 	}
-	s.RUnlock()
 
 	// Using a write lock, delete the expired sessions
 	s.Lock()
--- a/server/irmaserver/sessions_test.go
+++ b/server/irmaserver/sessions_test.go
@@ -1,0 +1,51 @@
+package irmaserver
+
+import (
+	irma "github.com/privacybydesign/irmago"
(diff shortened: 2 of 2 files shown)


ivard:
Looks good and it also works. I made a unit test to verify that it actually works. It is maybe an idea to include this unit test now we have it. Maybe in a `sessions_test.go` within `server/irmaserver`?

```
func TestMemoryStoreNoDeadlock(t *testing.T) {
	store := &memorySessionStore{
		requestor: make(map[string]*session),
		client:    make(map[string]*session),
		conf:      nil,
	}

	logger := logrus.New()
	logger.Level = logrus.FatalLevel
	s, err := New(&server.Configuration{Logger: logger})
	require.NoError(t, err)
	defer s.Stop()

	req, err := server.ParseSessionRequest(`{"request":{"@context":"https://irma.app/ld/request/disclosure/v2","context":"AQ==","nonce":"MtILupG0g0J23GNR1YtupQ==","devMode":true,"disclose":[[[{"type":"test.test.email.email","value":"example@example.com"}]]]}}`)
	require.NoError(t, err)
	session := s.newSession(irma.ActionDisclosing, req)
	store.add(session)

	session.Lock()
	deletingCompleted := false
	addingCompleted := false

	go func() {
		store.deleteExpired()
		deletingCompleted = true
	}()
	go func() {
		session2 := s.newSession(irma.ActionDisclosing, req)
		store.add(session2)
		addingCompleted = true
	}()

	// Check whether the IRMA server doesn't hang
	time.Sleep(100 * time.Millisecond)
	require.True(t, addingCompleted)
	require.False(t, deletingCompleted)
	
	// Check whether the deleting continues when the session completes
	session.Unlock()
	time.Sleep(100 * time.Millisecond)
	require.True(t, deletingCompleted)
}
```

Shall I add this?

sietseringers:
Good idea, yes. Two suggestions:

* Instead of manually creating a `memorySessionStore` you could I think also just say `store := s.sessions.(*memorySessionStore)`. Also, then you don't have to explicitly add the sessions to the `store` anymore since `newSession` already does that.
* Maybe also add some sleep time between the starting of the two goroutines, so that we can be sure that the first one is running when the second one starts.

ivard:
I've added the unit test. I chose to not initialise `session` manually, since `deleteExpired` uses quite a lot of the administration and otherwise I'm afraid the test will fail because the dummy session wasn't correct. 

I removed the separate `store.add` in the unit test, since `s.newSession` appears to do that under water already. Futhermore I put the unlocking in a `defer` to make sure the test itself doesn't deadlock.

ivard (approved):

PR #155, opened 2021-06-08
Fix SSE in combination with chained sessions

sietseringers:
The presence of a next session URL in the session status message informs the frontend that there is a next session, so that it should not yet tell the user the session has finished. Previously the session status was set to DONE before the next session URL was retrieved. When SSE is used and the session status changes, the frontend is notified of this almost immediately. Thus in this case, it did not receive the next session URL, considering the session (chain) finished.


ivard (changes requested):
The change in commit 46a49ab looks good and it works nicely. This branch only needs some rebasing to `master`. Currently the diff looks enormous.

sietseringers:
Done.

ivard (approved):

PR #157, opened 2021-06-08
bump fxamacker/cbor to v2.3.0, use secure options

x448:
Bump fxamacker/cbor to v2.3.0 and create internal/cbor/cbor.go to wrap CBOR functions using secure options. 

CBOR is encoded using Core Deterministic Encoding defined in RFC 8949, which obsoletes Canonical CBOR defined in RFC 7049.  Old decoders will still be able to decode newly encoded data.

- [x] bump fxamacker/cbor from v1.5 to v2.3 to benefit from bug fixes and speedups
 - [x] improve security of CBOR serialization by creating and using internal/cbor/cbor.go
     - [x] encoding -- use Core Deterministic Encoding with indefinite length items disabled
     - [x] decoding -- detect and reject duplicate map keys when decoding CBOR, reject indefinite length items
     
Limit on max array elements can be increased if they're too small.

```Go
// irmago/internal/cbor/cbor.go 
const MaxArrayElements = 1024 * 256 // this limit can be set as high as 2147483647
const MaxMapPairs = 1024 * 256      // this limit can be set as high as 2147483647
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,24 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.0] - 2021-12-17
+
+### Added
+
+* Support for [stateless IRMA server using Redis](https://irma.app/docs/stateless)
+* Added Dockerfile and docker-compose files for running `irma`, the unit tests, and/or the services required by the unit tests
+
+### Changes
+
+* Improve error messages of IRMA server in case of invalid session requests
+
+### Fixed
+
+* Fix panic when an issuance request contains a credential ID consisting of less than three parts
+* Ensure session handler callback function, when specified, is also called when session expires
+* Several small bugs in MyIRMA backend server
+
+
 ## [0.8.0] - 2021-07-27
 This release contains several large new features. In particular, the shoulder surf prevention feature brings a number of breaking changes in the API, mainly within the `irmaserver` package.
 
@@ -191,6 +209,7 @@ This release contains several large new features. In particular, the shoulder su
 - Combined issuance-disclosure requests with two schemes one of which has a keyshare server now work as expected
 - Various other bugfixes
 
+[0.9.0]: https://github.com/privacybydesign/irmago/compare/v0.8.0...v0.9.0
 [0.8.0]: https://github.com/privacybydesign/irmago/compare/v0.7.0...v0.8.0
 [0.7.0]: https://github.com/privacybydesign/irmago/compare/v0.6.1...v0.7.0
 [0.6.1]: https://github.com/privacybydesign/irmago/compare/v0.6.0...v0.6.1
--- a/version.go
+++ b/version.go
@@ -5,4 +5,4 @@
 package irma
 
 // Version of the IRMA command line and libraries
-const Version = "0.8.0"
+const Version = "0.9.0"


sietseringers:
Thanks!

We also use CBOR in a dependency of ours, so I've moved your code to there, see https://github.com/privacybydesign/gabi/pull/28. Once that is merged, we'll update this so that we use the new subpackage from there.

x448:
> We also use CBOR in a dependency of ours, so I've moved your code to there, see https://github.com/privacybydesign/gabi/pull/28. Once that is merged, we'll update this so that we use the new subpackage from there.

Sietse Ringers,

Moving my code that way excludes my name from irmago's contributors list, makes the number of lines I contributed to irmago get reported as zero, and makes you appear as the author or co-author of my work.

Doing that move on Christmas after 6 months of silence was unexpected.

sietseringers:
Yes, I understood that when you said the same thing [here](https://github.com/privacybydesign/gabi/pull/28#issuecomment-1001709556), and I apologized for it [here](https://github.com/privacybydesign/gabi/pull/28#issuecomment-1001719409).

The structure of `gabi` and `irmago` is such that this code fits better in `gabi` than here. Since our discussion there, I have been waiting for you to re-open another PR with these changes.

If you do that, there will be corresponding changes that need to be made here in `irmago` as well (bumping `gabi` and importing the new `gabi` subpackage). That way you will end up on the contributor list in both repositories.

x448:
Sietse Ringers,

Thank you for saying you understood the effects of your move on the author.  I wasn't sure because the apologies were generic "I apologize for the long silence, and for closing the other PR".

Until your reply here, there was no indication I would not remain excluded from irmago contributors list.  I didn't know if my security improvements to irmago would be buried by a `nothing-to-see` PR with a title like "moved some code to gabi".

This PR is showing `Commits 0` which permanently hides about 105 lines of Go code I added to irmago.  Are there any disadvantages to undoing the hiding and merging this PR before I proceed with opening new PRs in gabi and irmago?

sietseringers:
Well, yes; then the code would be here as well as in gabi; i.e. code duplicated. If there were no disadvantage to having your code here in `irmago`, I would not have closed this PR.

The reason why I closed this PR and opened the other one (which I shouldn't have done the way I did as you pointed out), moving this code to a (non-internal) subpackage of `gabi`, was so that we could have this code in one place instead of two. `irmago` and `gabi` are for most intents and purposes a single project that happens to be spread over two repositories, and both of them use `cbor`, so this code fits better there.

If you open a fresh `gabi` PR we can merge the code there, and then you can open a new `irmago` PR using the new `gabi` code with a name and description as you wish.

sietseringers:
By the way, I can't seem to undo the hiding of that commit (which was another mistake of mine). But since this is your PR, you yourself should be able to force-push your master branch back to commit `9514178`, which contained your code. This PR is going to stay closed, however, for the reasons I explained above.

x448:
> Well, yes; then the code would be here as well as in gabi; i.e. code duplicated. If there were no disadvantage to having your code here in irmago, I would not have closed this PR.
>
>The reason why I closed this PR and opened the other one (which I shouldn't have done the way I did as you pointed out), moving this code to a (non-internal) subpackage of gabi, was so that we could have this code in one place instead of two. 

Thank you for seeing the code reuse opportunities made easier by this PR.  This PR not only improves security of irmago but also makes code reuse easier.

My question more precisely was, are there any disadvantages to this sequence of steps:

1.  unhide the changes in this PR and merge this PR
2.  open PR in gabi to improve security of CBOR encoding and decoding
3.  open PR in irmago to use CBOR via gabi to reuse the improvements

Step 1 gives me credit for making security improvements and code reuse improvements to irmago on June 8, 2021.

This would avoid replacing my security improvements to irmago with a tiny `nothing-to-see` PR about reusing code in gabi.

sietseringers:
Alright then.

As I said in my previous comment, you'll have to do step 1 yourself; I can't do that since it concerns your fork of `irmago` (`git` says  `! [remote rejected]   master -> master (permission denied)` if I try).

You should be able to do this as follows in your `irmago` checkout:
```sh
git checkout master
git reset --hard 9514178
git push --force
```

sietseringers:
Can you do that?

Or what do you mean with that emoji on my previous post?

sietseringers:
Using an old backup of my `irmago` checkout, I have restored the contents of your original `9514178a` commit into a new branch called `cbor-subpackage`.

PR #161, opened 2021-06-21
[IRMA] Add Dockerfile for irma server

vandernorth:
Hi,

I wanted to run IRMA server in our Kubernetes cluster and was in need of a docker image. I can imagine others would like this too.
Not sure if there is a reason this was not yet there?

I also have code for the gitlab-ci file which creates a docker image and adds it to the gitlab containers repository if you are interested.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,27 @@
-FROM buildpack-deps:buster as build
+# Use variable base image, such that we can also build for other base images, like alpine.
+ARG BASE_IMAGE=debian:buster-slim
 
-WORKDIR /build
-RUN wget https://gitlab.science.ru.nl/irma/github-mirrors/irmago/-/jobs/artifacts/master/download?job=binaries -O artifact.zip -q
-RUN unzip -j artifact.zip
+FROM golang:1.16-buster as build
 
-FROM golang:1.16-alpine
+# Set build environment
+ENV GOOS=linux
+ENV GOARCH=amd64
+ENV CGO_ENABLED=0
 
-COPY --from=build /build/irma-master-linux-amd64 /usr/local/bin/irma
+# Build irma CLI tool
+COPY . /irmago
+WORKDIR /irmago
+RUN mkdir -p /build
+RUN go build -a -ldflags '-extldflags "-static"' -o "/build/irma" ./irma
 
-EXPOSE 8088
+FROM $BASE_IMAGE
 
-CMD [ "irma", "server"]
+# The debian image does not include openssl, so we have to install this first.
+RUN if which apt-get &> /dev/null; then apt-get update && apt-get install -y ca-certificates openssl; fi
+
+COPY --from=build /build/irma /usr/local/bin/irma
+
+# Include schemes in the Docker image to speed up the start-up time.
+RUN irma scheme download
+
+ENTRYPOINT ["irma"]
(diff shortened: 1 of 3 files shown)


ivard:
Nice, good point that it would be good to add a `Dockerfile`. Now you simply download the artifact from Gitlab, but I'd prefer to include the whole building process in the build stage of the `Dockerfile`. Then you can also make a `master` build using `docker build`. This would not be possible using Gitlab, because we don't make artifacts on Gitlab for every commit.

If I have some time left, I'll make a proposal.

ivard:
I updated it now to build from source and to use a variable build image, such that we can build for different base images. Then people can use the base image of their choice.

ivard (dismissed):
We have the policy that two people should review external PRs, especially because I changed almost everything :). Let's hope and see.

ivard:
Maybe it is an idea to add an image name, such that people can recognize it in their `docker image ls`. 

Ideally, that should be the same as the image name we want to publish it with, if we want to do that at some point. Maybe `privacybydesign/irmago`?

sietseringers:
Seems like a good idea. I'd say just `privacybydesign/irma` then, since it contains the `irma` binary.

Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Use variable base image, such that we can also build for other base images, like alpine.
+ARG BASE_IMAGE=debian:buster-slim
+
+FROM golang:1.16-buster as build
+
+# Set build environment
+ENV GOOS=linux
+ENV GOARCH=amd64
+ENV CGO_ENABLED=0
```
sietseringers, line 9: Aren't these lines superfluous? They all contain the default values (for this platform).

ivard: Yeah I thought so too, but if I use `alpine` as base image (which a lot of people do), then it complained about `CGO`. And I always use debian buster to build, so it is not related to the build process itself.

The `GOOS` and `GOARCH` can be removed I think indeed.

ivard: CGO is enabled by [default](https://pkg.go.dev/cmd/cgo). The `alpine` image is so small, that it does not have native bindings for CGO, so that causes the issues. Therefore, I think it is better to build without CGO, such that we get a binary that works everywhere (just like we do in the regular pipeline).

ivard (approved):

ivard (approved):
I used the `edge` tag to make clear that when people run this command without thinking, they don't build a latest release. When we make the first irmago release with the `Dockerfile` being present, we can add a comment in the README that they can check out the latest git tag to build a proper latest release.

sietseringers:
> I used the `edge` tag to make clear that when people run this command without thinking, they don't build a latest release. When we make the first irmago release with the `Dockerfile` being present, we can add a comment in the README that they can check out the latest git tag to build a proper latest release.

Why did you call this `edge`? Would it not be logical to call it `master` since it builds the `master` branch?

ivard:
Could also be. If you have a local feature branch, then it builds than one of course, so maybe `dev` is more generic then?

sietseringers:
Oh right, I didn't think of feature branches; then `master` makes little sense indeed. In that case `edge` is just as good as `dev` as far as I'm concerned, so we can keep it as it is.

ivard (dismissed):

ivard (approved):
Rebased the PR to `master`

sietseringers (approved):

PR #163, opened 2021-07-09
Use upstream cobra

sietseringers:
This drops our forks of `viper`, `cobra` and `pflag`, switching back to upstream.

Previously we depended on these forks for the following two reasons:
* Ensuring that options consisting of multiple words use dashes when used as a flag (e.g. `--static-path /foo/bar`) and underscores when used in a configuration file (`{"static_path": "/foo/bar"}`). 
* Inserting headers between flags in the help message of (a.o.) `irma server -h`.

These are now done as follows, respectively:
* Using [this mechanism](https://github.com/spf13/viper/issues/1044).
* Using a custom template for the help message set using `cobra.Command.SetUsageTemplate()`, which invokes a new function that inserts the headers.


go.mod
```diff
@@ -32,15 +31,13 @@ require (
 	github.com/onsi/gomega v1.9.0 // indirect
 	github.com/privacybydesign/gabi v0.0.0-20210520084001-626ad7db13a6
 	github.com/shopspring/decimal v1.2.0 // indirect
-	github.com/pelletier/go-toml v1.2.0 // indirect
-	github.com/sietseringers/cobra v1.0.1-0.20200909200314-c50c3838234b
 	github.com/sietseringers/go-sse v0.0.0-20200801161811-e2cf2c63ca50
```
synaptic-cleft, line 34: Any chance we could insource this, too? The original code (https://github.com/andrewstuart/go-sse) hasn't been updated for quite a while.

Have we ever contacted Andrew Stuart?

sietseringers: I have not.

I don't think it makes much sense to try to insource this. Of the original code of that repo, there is little left in my fork; I have even added a design goal that he apparently did not have (namely, adher to the SSE spec). The code is still fairly simple and small, however, so instead of rebasing onto changes from upstream (which don't appear to be coming), I was just planning to maintain my fork myself (if and when necessary).

synaptic-cleft (approved):

PR #172, opened 2021-09-18
Added instructions how to make Docker builds of a release

ivard:
Additional instructions on the proposal on #161 that can be introduced when we released a new irmago version with a `Dockerfile` included in the repo. Then people can simply build Docker images for release builds in a one-liner. Another advantage is that we don't necessarily have to publish the Docker image anymore. People can build it themselves without having to install additional tooling.

I already make this PR now, since I sorted it out this week. Then I can better immediately write it down, such that we don't forget it again. Otherwise, I keep wandering around with it.

It does not have to be merged right away. It is only useful when the `Dockerfile` is merged and released.


README.md
```diff
@@ -21,6 +27,20 @@ To install the `irma` command line tool:
 
     go install ./irma
 
+You can also include the `irma` command line tool in a Docker image, using a base image of your choice. The default base image is Debian's `buster-slim`.
+
+    docker build --build-arg BASE_IMAGE=alpine -t privacybydesign/irma:edge .
+
+When you build for production, we recommend you to build the [latest release](/releases/latest). You should replace `v0.0.0` with the latest version number.
+
+    docker build -t privacybydesign/irma https://github.com/privacybydesign/irmago#v0.0.0
```
ivard, line 36: It is a bit of a pity that there is no default `latest` tag or another way to choose the latest release.

sietseringers (approved):

PR #174, opened 2021-09-22
Fix panic when an issuance request contains a credential ID consisting of not enough parts

sietseringers:


ivard:
Makes sense.

Maybe we also want to add that the server explicitly gives a `500` if an unexpected error occurs. That can be done by enabling the [`Recoverer` middleware](https://pkg.go.dev/github.com/go-chi/chi/middleware#Recoverer). Then it at least always returns something instead of failing silently.

sietseringers:
That's a nice idea.

However, I would suggest doing that with a custom bit of middleware instead of [`Recoverer`](https://pkg.go.dev/github.com/go-chi/chi/middleware#Recoverer):

* `Recoverer` returns the string `Internal Server Error` as the response body, while we generally return JSON-serialized [`irma.RemoteError`](https://pkg.go.dev/github.com/privacybydesign/irmago#RemoteError) instances on errors;
* `Recoverer` uses a `chi` logger instead of the logger we use.

Given this, I think this merits its own PR.

PR #175, opened 2021-09-27
Added example configuration and run instructions for irma keyshare server

ivard:
PR on top of #161 .


ivard:
Needs to be rebased when #161 is merged.

testdata/configurations/keyshareserver.yml
```diff
@@ -0,0 +1,25 @@
...
+
+schemes_path: testdata/irma_configuration
+schemes_update: 0
+privkeys: testdata/privatekeys
+url: http://localhost:8080/
+
+keyshare_prefix: /irma_keyshare_server/api/v1/
+keyshare_attribute: test.test.mijnirma.email
+
+db_str: postgresql://testuser:testpassword@postgres:5432/test
```
ivard, line 11: We cannot use `localhost` in the urls due to a limitation in Docker on MacOS. Check the comment in the networks section of `docker-compose.yml`.

docker-compose.yml
```diff
@@ -1,19 +1,41 @@
...
     environment:
       POSTGRES_USER: testuser
       POSTGRES_PASSWORD: testpassword
       POSTGRES_DB: test
+    networks:
+      - irma-net
     ports:
       - 5432:5432
+  postgres-init:
+    image: postgres:13-buster
```
ivard, line 16: It would be good to upgrade to the new stable Debian release `bullseye`. That can be done later, because otherwise I would have to open a third Docker PR :)

ivard:
@sietseringers You'd like the solution to change the unit tests to not make use the `irma_keyshare_server` prefix better right? Because I can simply remove the commits implementing the `prefix` option from the top of this branch and go for that other solution. I don't have a strong preference for one of the two solutions.

sietseringers:
Yes, that's right. Sounds good.

ivard:
I now use port `8081` in the example configuration of the myirmaserver, because otherwise both servers would get the same URL in the `irma_configuration`. This is not necessarily a problem, but I thought it prevents some confusion in the future. I tested the myirma configuration too, and this one works too in the Docker set-up now.

docker-compose.yml
```diff
@@ -1,19 +1,41 @@
...
+    networks:
+      - irma-net
+    depends_on:
+      - postgres
+    volumes:
+      - ./server/keyshare/schema.sql:/schema.sql
+      - ./server/keyshare/cleanup.sql:/cleanup.sql
+    # We have to wait until the database is up and running.
+    # Database might already be running, so we might need to do a cleanup first.
+    command: /bin/sh -c "sleep 5 && (psql -f schema.sql || (psql -f cleanup.sql && psql -f schema.sql))"
```
sietseringers, line 31: Couldn't we instead do [`DROP SCHEMA IF EXISTS irma CASCADE;`](https://www.postgresql.org/docs/8.2/sql-dropschema.html), and then simplify this to the following?

```sh
/bin/sh -c "sleep 5 && psql -f cleanup.sql && psql -f schema.sql"
```

testdata/configurations/keyshareserver.yml
sietseringers: This means that after this change, this file as well as `myirmaserver.yml` will work only if postgreSQL and Mailhog have been started using Docker compose, and not anymore when they run locally as daemons. Is there no other way that keep both options working?

sietseringers (changes requested):

docker-compose.yml
ivard: Yes we can, I'd propose to change this in the `cleanup.sql` script then. Otherwise we duplicate this logic.

ivard:
The default options do not work, because exposing ports when using host networking is not supported in Docker Desktop (on Windows and MacOS) and the DNS alias `host.docker.internal` is not supported on Linux.

I tried some things, and the only work-around that seems to work on both Linux and MacOS is to use `postgres.localhost` and `mailhog.localhost` as domain and add those as alias to the postgres and mailhog service. When running it natively, those urls are resolved as `127.0.0.1`, so they'll just work. And in the Docker container, the IP will be overrulled in the `/etc/hosts` to actually use the IP of the other Docker container.

That is the only thing I can think of that works on all platforms without big problems. Would such a thing make you more happy?

sietseringers:
> the only work-around that seems to work on both Linux and MacOS is to use postgres.localhost and mailhog.localhost as domain and add those as alias to the postgres and mailhog service. When running it natively on the host, those urls are resolved as 127.0.0.1, so they'll just work.

Turns out that this depends on your DNS server: some do resolve this to 127.0.0.1, some don't. However, this is apparently the best we can get, so I think we should do it like this, yes.

ivard:
You mean like keeping it this way with `mailhog` and `posgres` domain names and then add in the documentation that people that people that run it without Docker need to manually change their `/etc/hosts` file or change the urls in the configuration?

I mean if the localhost hack does not work in all cases, then I would be in favour of not choosing the hacky solution and simply add some documentation. Most people will probably use Docker, because this is how it is described in the README now. So specifying a work-around for people who want to do it themselves, without Docker, seems acceptable to me.

sietseringers:
> You mean like keeping it this way with mailhog and posgres domain names and then add in the documentation that people that people that run it without Docker need to manually change their /etc/hosts file or change the urls in the configuration?

No, I meant use `postgres.localhost` and `mailhog.localhost`, since that always works for Docker and sometimes works locally, so of all options it achieves the most. If Docker accepts domains looking like that, this doesn't seem all that hacky to me. Additionally, I didn't mention documenting that people modify their `/etc/hosts`; that would be a rather harsh requirement. I only meant that that is a way of making it also work locally, for me personally.

ivard:
I added the commits now, so then can you test whether the `localhost` hack works for you sufficiently.

sietseringers (approved):

PR #180, opened 2021-10-08
Ensure session handler callback function is also called on expired sessions

sietseringers:
This required moving the callback bookkeeping from the server instance to the session itself,
since the session store has no direct access to the server instance. However, keeping these
on the sessions themselves seems more logical anyway.

### DIFF
--- a/server/irmaserver/sessions.go
+++ b/server/irmaserver/sessions.go
@@ -139,10 +139,6 @@ func (s *memorySessionStore) deleteExpired() {
 				s.conf.Logger.WithFields(logrus.Fields{"session": session.requestorToken}).Infof("Session expired")
 				session.markAlive()
 				session.setStatus(irma.ServerStatusTimeout)
-				if handler := session.handler; handler != nil {
-					session.handler = nil
-					go handler(session.result)
-				}
 			} else {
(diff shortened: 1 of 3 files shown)


server/irmaserver/sessions.go
```diff
@@ -138,6 +139,10 @@ func (s *memorySessionStore) deleteExpired() {
 				s.conf.Logger.WithFields(logrus.Fields{"session": session.requestorToken}).Infof("Session expired")
 				session.markAlive()
 				session.setStatus(irma.ServerStatusTimeout)
+				if handler := session.handler; handler != nil {
+					session.handler = nil
+					go handler(session.result)
+				}
```
ivard, line 145: I saw that `session.setStatus` is also called to process cancellation. Cancellation can also be done using the `CancelSession` API function. This function is for example used by the `requestorserver` to cancel a session. When using the API function, the `sessionMiddleware` of `irmaserver` is also bypassed, so this issue will occur in that case too right?

I think that variant of the same bug should be resolved too then.

ivard (changes requested):
The bug fix looks good, but I think we forgot to fix it on another place where the same bug occurs (during cancellation). Check the comment.

PR #182, opened 2021-10-12
Ensure wizard items order is always stable

sietseringers:
The dependency graph of wizard items does not always determine a unique order in which the items are to be executed, and indeed items can switch places within a single wizard across invocations. This is now fixed.

### DIFF
--- a/descriptions.go
+++ b/descriptions.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"path/filepath"
 	"regexp"
+	"sort"
+	"strings"
 
 	"github.com/go-errors/errors"
 	"github.com/privacybydesign/irmago/internal/common"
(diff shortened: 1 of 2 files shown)


descriptions.go
```diff
@@ -262,7 +264,14 @@ func buildDependencyTree(contents []IssueWizardItem, conf *Configuration, credsm
 	var result []IssueWizardItem                         // to return
 	resultMap := map[CredentialTypeIdentifier]struct{}{} // to keep track of credentials already put in the result slice
 	for i := len(depTree) - 1; i >= 0; i-- {
+		var credIDs []CredentialTypeIdentifier
```
synaptic-cleft, line 267: You've added order to the wizards now. Currently a test sometimes fails because the test assumes order. Is order really necessary? Couldn't we better change the test not to depend on the order instead of changing production code?

PR #184, opened 2021-10-29
Added Github action for linter

ivard:
`golint` is not included on purpose, because we have too many issues there still.

For the time being, the status checks also don't run `gocylco` (for code complexity). The configuration of that is a bit more complex, because we have to only check the diff here, and I was not able to get that fully right. In some cases some code complexity might be unavoidable, so it might also be good to add output the `gocyclo` output to a PR review, such that it can be manually judged. This is possible, but requires some additional configuration too.

### DIFF
--- a/.github/workflows/status-checks.yml
+++ b/.github/workflows/status-checks.yml
@@ -1,0 +1,74 @@
+# Workflow to check whether changes to master fulfill all requirements.
+name: Status checks
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Go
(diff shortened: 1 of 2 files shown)


.github/workflows/status-checks.yml
```diff
@@ -0,0 +1,74 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.16
+
+    - name: Run gofmt
+      run: gofmt -d -e .
```
ivard, line 51: Go report card uses the simplified option of `gofmt`. Maybe we want to use this option too. Then we would have some issues to fix.
https://goreportcard.com/report/github.com/privacybydesign/irmago#gofmt

sietseringers: I suggest doing this after the session test refactor has been done, to avoid a merge conflict.

sietseringers (approved):

PR #187, opened 2021-11-17
Feat: added option to specify which TLS configuration to use for outbound connections

ivard:
For `irmamobile` we need `irmago` to use additional root certificates, besides the ones loaded in the system. In the current setup, irmamobile would need to write the certificates from memory to disk, such that irmago can load them from the location specified by the environment variable `SSL_CERT_DIR`. This creates superfluous overhead.

Therefore, it would be nice to have support to immediately include the certificates in Go. Then we don't need to write them to disk first.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -3,6 +3,7 @@ package irma
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/hex"
 	"encoding/json"
 	"io"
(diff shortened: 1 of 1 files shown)


transport.go
```diff
@@ -60,6 +63,12 @@ func SetLogger(logger *logrus.Logger) {
 	sseclient.Logger = log.New(Logger.WithField("type", "sseclient").WriterLevel(logrus.TraceLevel), "", 0)
 }
 
+// SetTLSClientConfig sets the TLS configuration being used for future outbound connections.
+// A TLS configuration instance should not be modified after being set.
+func SetTLSClientConfig(config *tls.Config) {
+	tlsClientConfig = config
+}
```
ivard, line 70: Having this support, it would be nice to add a unit test with retryable HTTP in combination with an IRMA server with TLS being enabled. This is easier when having the unit test refactors of Redis, so i'd propose to not include that here.

PR #189, opened 2021-12-10
Replace dgrijalva/jwt-go with golang-jwt/jwt

### LINKED ISSUE #188: Replace jwt-go with golang-jwt
jwt-go is deprecated and contains a security problem issue: https://github.com/advisories/GHSA-w73w-5m7g-f7qc.
The advice is to replace it with the community supported https://github.com/golang-jwt/jwt.

stevenvegt:
Fixes #188


sietseringers:
Thanks! The issue in `jwt-go` was not a problem for us because we don't use the `aud` field, but this is of course better.

sietseringers (approved):

synaptic-cleft (approved):

PR #194, opened 2021-12-22
Improve scheme language validation

sietseringers:
Schemes and items contained in schemes can now define the languages into which their texts should be translated. This provides a way to with languages other than English and Dutch (which until now were the only languages we supported), and with scheme items that are not expected to be translated in Dutch. It also provides a way to disable language checking, when it is not needed.

### Design
The following structs have gained a `Languages []string` field:
* Issuer schemes: `SchemeManager`, `Issuer`, `CredentialType`, denoted as follows in XML:
  ```xml
  <Languages>
    <Language>en</Language>
    <Language>nl</Language>
  </Languages>
  ```
* Requestor schemes: `RequestorScheme`, `RequestorInfo`, `IssueWizard`, denoted as follows in JSON:
  ```json
  "languages": ["en", "nl"]
  ```

During scheme parsing and validation, for each translatable item it is checked that each expected language is present. If not, a warning is emitted.

* If a `CredentialType` has no `Languages` set, then those of its `Issuer` apply. If an `Issuer` has no `Languages` set, then those of its `SchemeManager` apply. Similarly for `IssueWizards`, `RequestorInfo` and `RequestorSchemes`.
* If an instance of one of these structs has no languages defined at all, and neither do the structures they are contained in, then no language validation is done at all. This enables a way to disable language validation in demo schemes.
* This feature is backwards compatible in the usual way: older IRMA app or `irma` instances automatically ignore these new JSON and XML fields when unmarshaling.


descriptions.go
```diff
@@ -83,11 +86,11 @@ type CredentialType struct {
...
-	Category            TranslatedString
-	FAQIntro            TranslatedString
-	FAQPurpose          TranslatedString
-	FAQContent          TranslatedString
-	FAQHowto            TranslatedString
+	Category            *TranslatedString
+	FAQIntro            *TranslatedString
+	FAQPurpose          *TranslatedString
+	FAQContent          *TranslatedString
+	FAQHowto            *TranslatedString
```
ivard, line 93: After this change we don't have the requirement anymore that a translation is present for every language, right? So why are all these pointers still needed? In the case where there is no translation, you could simply assign the empty map. `omitEmpty` also omits the empty map, so on that point there is no difference.

descriptions.go
```diff
@@ -186,6 +192,8 @@ type (
 		SessionURL *string                   `json:"sessionUrl,omitempty"`
 		URL        *TranslatedString         `json:"url,omitempty"`
 		InApp      *bool                     `json:"inapp,omitempty"`
+
+		languages []string
```
ivard, line 196: It looks a bit weird to me to add the languages separately to the struct, while the languages should already be present in all `TranslatedString`s. It seems more logical to me to give this `languages` as a parameter to `Validate`, or to add a check in the `Validate` function of `IssueWizard` that all its `IssueWizardItem`s contain all expected languages.

descriptions.go
sietseringers: (Terminology: given a credential type, let us denote with "specified languages" the languages specified in the `<Languages>` tag of that credential type, *or* its issuer if the credential type does not have that tag, *or* its scheme if the issuer does not have that tag.)

These are fields for which it is allowed that they are empty regardless of the specified languages. That's what the pointer specifies:
* Fields of type `TranslatedString` (e.g. the `<Name>` of a credential type) must always contain all of the specified languages,
* Fields of type `*TranslatedString` (such as these ones) may have the null pointer as value regardless of the specified languages.

ivard: Yes that's what I meant indeed. Besides how it's printed in XML or JSON, there is no real difference anymore between a field not being present, and the empty map right. Or is there a slightly different meaning?

Because now you need to check at all places both whether or not the field is `nil` and whether the particular language is available. Only checking the latter could also satisfy if there is no different meaning.

descriptions.go
sietseringers: We could do that, but it would create an inconsistency: a lot of other structs that contain translatable strings have a `lanuages []string` like this one. Namely, the `SchemeManager`, `Issuer`, `CredentialType`, `RequestorScheme`, `RequestorInfo`, and `IssueWizard` structs. Additionally, none of the other `validate` functions (except the one of `TranslatedString` which is special anyway) take languages as parameter.

As to:
> or to add a check in the `Validate` function of `IssueWizard` that all its `IssueWizardItems` contain all expected languages.

We could do that, but it seems cleaner to me to leave this check to the `validate` function of the `IssueWizardItem` itself.

descriptions.go
sietseringers: The point of this pointer is the following: some fields must always be translated (according to the specified languages), and some fields may always be empty. We want some way to encode that distinction using the type system, and a pointer seems a fine choice to me for that. It's the same as `int`, which always has some value, and `*int`, where `null` indicates absence.

> there is no real difference anymore between a field not being present, and the empty map right

Well, I would say there is a difference: a nil map is not the same as an empty map. The former may be fine (or not, depending on the field), but the latter is strange: if you meant to not specify translations, `nil` would be more explicit. An empty map to me looks like you forgot to fill it. So I would say, you either say `nil`, or you give a non-empty map.

sietseringers: Correspondingly, taking e.g. `FAQIntro`, in XML you either say nothing or you specify an `<FAQIntro>` containing at least one language. That's what this code currently enforces.

ivard: That would make sense, but I don't see any check doing this. When an empty `TranslatedValue` is passed and no languages are required, then this would be accepted now I think. Or do I overlook a check?

descriptions.go
ivard: That's a good point. It's both not very ideal. So let's go for consistency then.

descriptions.go
sietseringers: * In case of `*TranslatedString`, [here](https://github.com/privacybydesign/irmago/blob/improve-scheme-language-validation/irmaconfig.go#L775) we allow it to be nil;
* If it is not nil, then [a few lines below](https://github.com/privacybydesign/irmago/blob/improve-scheme-language-validation/irmaconfig.go#L783) we invoke the function that checks that it contains the required languages. In particular, if the map is empty then this function will complain (because an empty map does not contain the required languages).

sietseringers: On the other hand, if no languages are defined, then indeed that would be accepted. But that to me is one of the features of this PR. It will solve all the senseless warnings when you run `irma scheme verify` on `irma-demo`. And if no languages are defined, what languages could we enforce to be present? 🙂

ivard: So you say here

> An empty map to me looks like you forgot to fill it.

But currently, we allow this situation to happen in the XML. That's a bit strange then right?

Maybe another question, do we enforce `Languages` to be defined for a `SchemeManager`? If that is the case, then we already prevent the case that we don't know which languages we could expect.

sietseringers: Ah, now I finally understand what you mean 🙂 Sorry that took a while. We do allow that to happen, yes, but only when no languages are defined (which is by itself allowed). But even then it makes no sense to allow that, that's a good point. I have fixed it.

irmaconfig.go
```diff
@@ -779,6 +775,10 @@ func (conf *Configuration) validateTranslations(file string, o interface{}, lang
 			val = field.Interface().(TranslatedString)
 		}
 
+		if val != nil && len(val) == 0 {
+			conf.Warnings = append(conf.Warnings, fmt.Sprintf("%s has empty <%s> tag", file, name))
+		}
```
ivard, line 780: Maybe a small detail, but `val` can never be `nil` here if I'm right.

ivard (approved):

PR #201, opened 2022-01-19
Update viper/cobra to remove dgrijalva/jwt-go dependency

dmeijboom:
Seems like viper version `v1.7.0` (which `irmago` currently uses) still depends on the unmaintained `github.com/dgrijalva/jwt-go` library.

Changes:
- Updated `github.com/spf13/viper` to v1.10.1
- Updated `github.com/spf13/cobra` to v1.3.0


ivard (approved):
Thank you for the contribution.

Seems like a good idea, and all tests succeed so apparently it works.

sietseringers (approved):

PR #203, opened 2022-01-26
Feat: added option for static sessions and session packages to irma session command

ivard:
Useful for testing [static QRs](https://irma.app/docs/irma-server/#static-irma-qrs) and to handle sessions for the UniqueID issuer.

## Usage
    irma session --url $URL --static mystaticsession
    irma session --from-package $(curl "https://angrygames.nl/start_session.php?type=18plus") --pairing

### DIFF
--- a/irma/cmd/session.go
+++ b/irma/cmd/session.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"bufio"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"os"
@@ -43,7 +44,8 @@ irma session --sign irma-demo.MijnOverheid.root.BSN --message message
 irma session --issue irma-demo.MijnOverheid.ageLower=yes,yes,yes,no --disclose irma-demo.MijnOverheid.root.BSN
 irma session --request '{"type":"disclosing","content":[{"label":"BSN","attributes":["irma-demo.MijnOverheid.root.BSN"]}]}'
 irma session --server http://localhost:8088 --authmethod token --key mytoken --disclose irma-demo.MijnOverheid.root.BSN
-irma session --server http://localhost:8088 --static mystaticsession`,
+irma session --server http://localhost:8088 --static mystaticsession
+irma session --from-package '{"sessionPtr": ... , "frontendRequest": ...}'`,
 	Run: func(cmd *cobra.Command, args []string) {
 		var (
 			request    irma.RequestorRequest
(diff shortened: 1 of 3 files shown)


irma/cmd/session.go
```diff
@@ -42,42 +42,65 @@ with --message flags, or it can be specified as JSON to the --request flag.`,
...
-		flags := cmd.Flags()
-		disableUpdate, _ := flags.GetBool("disable-schemes-update")
-		if !disableUpdate {
-			if err = irmaconfig.UpdateSchemes(); err != nil {
-				die("failed updating schemes", err)
+		if staticFlag := flags.Lookup("static"); staticFlag.Changed {
+			if serverURL == "" {
+				die("Failed to read configuration", errors.New("--static must be combined with --server"))
+			}
+			pkg, err = staticRequest(serverURL, staticFlag.Value.String(), authMethod, key)
```
ivard, line 71: I have to look at this again, because propably static sessions only give a `sessionPtr` instead of a full session package.

irma/cmd/session.go
```diff
@@ -42,42 +44,69 @@ with --message flags, or it can be specified as JSON to the --request flag.`,
 irma session --sign irma-demo.MijnOverheid.root.BSN --message message
 irma session --issue irma-demo.MijnOverheid.ageLower=yes,yes,yes,no --disclose irma-demo.MijnOverheid.root.BSN
 irma session --request '{"type":"disclosing","content":[{"label":"BSN","attributes":["irma-demo.MijnOverheid.root.BSN"]}]}'
-irma session --server http://localhost:8088 --authmethod token --key mytoken --disclose irma-demo.MijnOverheid.root.BSN`,
+irma session --server http://localhost:8088 --authmethod token --key mytoken --disclose irma-demo.MijnOverheid.root.BSN
+irma session --url http://localhost:8088 --static mystaticsession
```
ivard, line 48: I use the --url option now, whereas normally the --server option would be logical here. But a user should enter the external url to make sure the IRMA app can reach it. This means the `--url` parameter description better matches, but this `--url` option is normally only used when running in library mode (so when not talking to another irma server instance).

It's both a bit ugly. What would you prefer @synaptic-cleft consistency in explanation, or consistency in which option is used when?

ivard: Or I should make a dummy call to the`--server` to check which external url is configured, such that we can use that one in the static QR. That would be best, but that is also a bit weird.

synaptic-cleft: I probably would have used server here, yeah, but I don't have a clear preference. I don't mind if it stays how you currently implemented it.

Maybe @sietseringers has a more pronounced opinion here.

synaptic-cleft (approved):

irma/cmd/session.go
```diff
@@ -230,70 +254,97 @@ func serverRequest(
 	}()
 
 	wg.Wait()
-	if err != nil {
-		return nil, err
+
+	if err == nil && pkg.Token != "" {
```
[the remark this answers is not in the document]
ivard: I'll refactor it to remove the nested if statements.

irma/cmd/session.go
```diff
@@ -230,70 +254,97 @@ func serverRequest(
...
+		result := &server.SessionResult{}
+		urlParts := strings.Split(pkg.SessionPtr.URL, "/irma/session/")
+		if len(urlParts) <= 1 {
+			// Custom URL structure is used, so we cannot determine result endpoint location.
+			return nil, nil
+		}
+		baseURL := strings.Join(urlParts[:len(urlParts)-1], "")
```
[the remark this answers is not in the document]
ivard: Yeah that was a mistake. I'll fix it.

irma/cmd/session.go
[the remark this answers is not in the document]
ivard: It should be better now.

irma/cmd/session.go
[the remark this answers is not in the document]
ivard: I fixed it.

irma/cmd/session.go
sietseringers: I agree that neither `--server` nor `--url` is a perfect match here. However, those two roughly correspond to the current two modes `irma session` supports (external server vs internal library), while the new static functionality is different from both of those.

Would it not be better to choose neither one, and introduce a new flag, e.g. `--static-server`? This allows us to write a clear parameter description for that new parameter, and it means that the `irma server` command will have three distinct options (`--url`, `--server`, `--static-server`) for each of its three modes (library, external server, static sessions).

ivard: Maybe that's the best solution indeed. A bit more verbose but then we can get the explanations clear. I'll update that.

synaptic-cleft (approved):

PR #206, opened 2022-01-31
Test: remove part of the hardcoded network ports

ivard:
The test sometimes fail because one of the hardcoded ports is not available. This PR removes most of these hardcoded ports.

The ones I did not remove where reliant on hardcoded URLs in the test configuration:
- Revocation server
- Keyshare server
- MyIRMA server
- Test server that hosts the schemes

I did also not remove the global variables for the URL of the requestor server, because the `optionReuseServer` of the `doSession` helper relies on a globally specified URL. I managed to make the URL being generated dynamically, but an improvement would still be to refactor `doSession` such that also the hardcoded `requestorServerPort` and `requestorServerURL` can be removed.

### DIFF
--- a/internal/sessiontest/helper_servers_test.go
+++ b/internal/sessiontest/helper_servers_test.go
@@ -222,7 +222,6 @@ func IrmaServerConfiguration() *server.Configuration {
 			revKeyshareTestCred: {RevocationServerURL: revocationServerURL},
 		},
 		JwtPrivateKeyFile: jwtPrivkeyPath,
-		StaticSessions:    map[string]interface{}{},
 	}
 }
 
--- a/internal/sessiontest/helper_servers_test.go
+++ b/internal/sessiontest/helper_servers_test.go
@@ -5,8 +5,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
+	"net/http/httptest"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
@@ -37,22 +40,18 @@ var (
 	jwtPrivkeyPath = filepath.Join(testdata, "jwtkeys", "sk.pem")
 )
 
+// Some urls are hardcoded in the test configuration, so we have to hardcode them here too.
 const (
-	irmaServerPort = 48680
-
 	schemeServerURL = "http://localhost:48681"
-
-	requestorServerPort = 48682
-	requestorServerURL  = "http://localhost:48682"
 
 	revocationServerPort = 48683
 	revocationServerURL  = "http://localhost:48683"
+)
 
-	staticSessionServerPort = 48685
-	staticSessionServerURL  = "http://localhost:48685"
-
-	nextSessionServerPort = 48686
-	nextSessionServerURL  = "http://localhost:48686"
+// The doSession helper expects the requestor server URL to be globally defined, to support the optionReuseServer.
+var (
+	requestorServerPort = findFreePort()
+	requestorServerURL  = fmt.Sprintf("http://localhost:%d", requestorServerPort)
 )
 
 type IrmaServer struct {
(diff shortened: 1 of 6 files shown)


internal/sessiontest/helper_servers_test.go
```diff
@@ -103,19 +108,19 @@ func StartIrmaServer(t *testing.T, conf *server.Configuration) *IrmaServer {
 		conf = IrmaServerConfiguration()
 	}
 
+	mux := http.NewServeMux()
+	httpServer := httptest.NewServer(mux)
+
+	// Make sure domain is used instead of IP address.
+	conf.URL = strings.Replace(httpServer.URL, "127.0.0.1", "localhost", 1)
```
sietseringers, line 115: Why is this necessary? If it is because IPv6 might be used, it would be helpful to add that to the comment above.

internal/sessiontest/helper_servers_test.go
```diff
@@ -229,19 +222,7 @@ func IrmaServerConfiguration() *server.Configuration {
...
-				},
-				Request: &irma.DisclosureRequest{
-					BaseRequest: irma.BaseRequest{LDContext: irma.LDContextDisclosureRequest},
-					Disclose: irma.AttributeConDisCon{
-						{{irma.NewAttributeRequest("irma-demo.RU.studentCard.level")}},
-					},
-				},
-			},
-		},
+		StaticSessions:    map[string]interface{}{},
```
sietseringers, line 225: Why is this line here? Can we not just remove it, and instantiate a new `map[string]interface{}` in `testStaticQRSession()`?

internal/sessiontest/redis_test.go
```diff
@@ -240,14 +240,16 @@ func TestRedisRedundancy(t *testing.T) {
...
+	ports := make([]int, 3)
 	servers := make([]*requestorserver.Server, len(ports))
-
-	for i, port := range ports {
+	for i := range ports {
+		port := findFreePort()
 		c := redisRequestorConfigDecorator(mr, cert, "", RequestorServerAuthConfiguration)()
-		c.Configuration.URL = fmt.Sprintf("http://localhost:%d/irma", port)
+		// Make sure URL of load balancer is being used in QRs.
+		c.URL = requestorServerURL
```
sietseringers, line 249: This changes the test behaviour itself, instead of just some ports. Strange that it wasn't already like this.

sietseringers (commented):
Nice! I have some small questions and remarks.

internal/sessiontest/helper_servers_test.go
ivard: Hmm, I didn't think of IPv6. The httptest library gives us an IP address instead of a domain. This causes the requestor name test to fail, because this one relies on the domain being `localhost`. I'm not sure whether we can rewrite that test in such a way that it expects an IP address. If that's not possible, we can use `url.Parse` to actually replace the host part with `localhost`. Then it should work in all cases.

internal/sessiontest/helper_servers_test.go
ivard: That's possible too indeed. I'll change it.

internal/sessiontest/redis_test.go
ivard: True, it was a mistake in the test I think. We overlooked that one. Luckily, it works like a charm, so it only was a testing mistake and not a functional mistake.

internal/sessiontest/helper_servers_test.go
ivard: Switching the domain in the `test-requestors` scheme to a IP address does not seem to be a very smart solution, because then we hardcode the use of IPv4 even more. I'd say let's do it this way then:
`
conf.URL = fmt.Sprintf("http://localhost:%d", portFromTestServer(httpServer))
`

With portFromTestServer being:
`
func portFromTestServer(s *httptest.Server) int {
	return s.Listener.Addr().(*net.TCPAddr).Port
}
`
This function can also be used in the function `findFreePort` to prevent code duplication. The `findFreePort` function is a bit of a hack for the `RequestorServer` now. Ideally, that function shouldn't be needed.

PR #209, opened 2022-02-25
Fix chained session deadlock

sietseringers:
If the next session URL responds so slowly that a polling lands in the middle, then:
1) the poll HTTP handler locks the session store, and first tries to lock the session before unlocking the session store;
2) the /proofs HTTP handler has the session locked, and tries locking the session store when the next session URL finally responds. But 1) still has the sessionstore lock, so 2) cannot complete and unlock the session lock, so 1) cannot complete and release the session store lock. Deadlock.

The fix is to make the unlocking of the memory session store in 1) immediate and not conditional on locking the session.

This PR also ensures that 4xx or 5xx responses result in warnings, regardless of the enabled logging level.

### DIFF
--- a/internal/sessiontest/helper_servers_test.go
+++ b/internal/sessiontest/helper_servers_test.go
@@ -185,7 +185,7 @@ func chainedServerHandler(t *testing.T, jwtPubKey *rsa.PublicKey) http.Handler {
 		})
 		require.NoError(t, err)
 
-		// give polling time to land in between
+		// Simulate a slowly responding server
 		time.Sleep(200 * time.Millisecond)
 
 		logger.Trace("2nd request: ", string(bts))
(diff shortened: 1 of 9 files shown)


internal/sessiontest/helper_servers_test.go
```diff
@@ -185,6 +185,9 @@ func chainedServerHandler(t *testing.T, jwtPubKey *rsa.PublicKey) http.Handler {
 		})
 		require.NoError(t, err)
 
+		// give polling time to land in between
```
synaptic-cleft, line 188: I like that there is a comment. I understand the comment with the current bug in my mind. However, I think why we really have this sleep is because it makes the integration test more realistic. It's a REST call. REST calls cost time. Therefore we wait here before we respond. I'd rather read this in the comment here than one specific use-case that triggered us to add the waiting time.

synaptic-cleft (approved):

PR #210, opened 2022-03-04
Use ordinary struct for unencrypted user secrets in keysharecore

sietseringers:


internal/keysharecore/usersecrets.go
```diff
@@ -31,69 +30,83 @@ var (
...
 }
 
 func (s *unencryptedUserSecrets) setKeyshareSecret(val *big.Int) error {
 	if val.Sign() == -1 {
 		return ErrKeyshareSecretNegative
 	}
 
-	data := val.Bytes()
-	if len(data) > 64 {
+	if val.BitLen()/8 > 64 {
```
synaptic-cleft, line 51: are you sure that this is what you want here? `.Bytes()` really got you the amount of bytes needed to fit all bits. `.BitLen()/8` might return 1 byte too little.

Mini example:
https://go.dev/play/p/IFRJ3T995v0

sietseringers: Well spotted, thanks!

synaptic-cleft (approved):

Issue #57, opened 2019-06-19
Use thread safe alternative of gocron


stevenvegt:
When embedding irmago in my go project, running test with the `-race` flag resuls in errors because the `gocron` library is not thread safe. See this issue: https://github.com/jasonlvhit/gocron/pull/60

There is a fork https://github.com/marcsantiago/gocron which contain all the fixes. It seems that the original gocron is no longer maintained. Will you consider using the fork or even a completely different package so we can use the `-race` flag again?

PR #219, opened 2022-04-22
Upgrade cron library to maintained go-co-op/gocron

reinkrul:
The current scheduler library (https://github.com/jasonlvhit/gocron) has been abandoned and causes race conditions (https://github.com/privacybydesign/irmago/issues/57). There is an active fork from the maintainers that inherited the original project, which is an (almost) drop-in replacement: https://github.com/go-co-op/gocron

This PR replaces jasonlvhit/gocron with go-co-op/gocron

### DIFF
--- a/internal/sessiontest/revocation_test.go
+++ b/internal/sessiontest/revocation_test.go
@@ -200,6 +200,8 @@ func TestRevocationAll(t *testing.T) {
 
 		// run scheduled update of accumulator, triggering a POST to our IRMA server
 		runAllSchedulerJobs(revServer.conf.IrmaConfiguration.Scheduler)
+		// give HTTP request time to be processed
+		time.Sleep(100 * time.Millisecond)
 
 		// check that both the revocation server's and our IRMA server's configuration
 		// agree on the same accumulator which has the same index but updated time
@@ -348,6 +350,8 @@ func TestRevocationAll(t *testing.T) {
 		accindex := sacc.Accumulator.Index
 		sacctime := sacc.Accumulator.Time
 
+		// wait for a moment to assure the accumulator's timestamp will actually differ
+		time.Sleep(time.Second)
 		// trigger time update and update accumulator
 		runAllSchedulerJobs(revServer.conf.IrmaConfiguration.Scheduler)
 
--- a/internal/sessiontest/revocation_test.go
+++ b/internal/sessiontest/revocation_test.go
@@ -1,3 +1,4 @@
+//go:build !local_tests
 // +build !local_tests
 
 package sessiontest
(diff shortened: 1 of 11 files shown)


sietseringers:
Thanks!

There are a few small issues with the PR, due to differences between the previous `gocron` and this one:

* The `Every()` method no longer accepts `uint64` parameters, which `irmago` uses at several invocations;
* The `Do()` method now returns an error, which complains about the above, but since the `Do()` method of the previous `gocron` did not return an error, this error is now ignored. This is why several of the unit tests failed. Ideally, at all invocations of the `Do()` method the error should be checked.

Additionally, when I build it a few changes are made to `go.mod` and `go.sum`.

Would you like to fix these issues? If not, we'd be happy to do that ourselves.

reinkrul:
> Would you like to fix these issues? If not, we'd be happy to do that ourselves.

I was just investigating these myself and am halfway fixing the same issues you found. I can probably push those fixes somewhere this week.

reinkrul:
I had to write some hacky code to wait for the scheduler jobs to finish when running unit tests. Otherwise the tests would fail on my development system (too slow on a Macbook?), when I increased the `Sleep()`'s here and there it mostly worked. The code I introduced to run all scheduler jobs and wait for them to finish might be a bit of a better solution than the sleeps, but still not very good.

I made a feature request in the scheduler repository to have it properly support unit testing; https://github.com/go-co-op/gocron/issues/326

internal/sessiontest/revocation_test.go
```diff
@@ -348,8 +349,8 @@ func TestRevocationAll(t *testing.T) {
 		sacctime := sacc.Accumulator.Time
 
 		// trigger time update and update accumulator
-		time.Sleep(time.Second)
```
sietseringers, line 351: This line causes a timestamp to be updated, which is verified below using `require.NotEqual(t, sacctime, sacc.Accumulator.Time)`. The removal of this line causes this to fail, can you reinstate it?

internal/sessiontest/revocation_test.go
```diff
@@ -196,9 +199,7 @@ func TestRevocationAll(t *testing.T) {
 		time.Sleep(time.Second)
 
 		// run scheduled update of accumulator, triggering a POST to our IRMA server
-		revServer.conf.IrmaConfiguration.Scheduler.RunAll()
-		// give request time to be processed
-		time.Sleep(100 * time.Millisecond)
```
sietseringers, line 201: This line should be reinstated as well; its purpose was not to wait for the scheduler to finish its jobs, but instead to give time to finish running to a HTTP handler which is invoked by a job.

sietseringers (changes requested):
Hi, sorry this has been quiet for so long. I figured out what caused the tests to fail, see the comments. After this, the PR should be fine to merge. Thanks!

reinkrul:
I must admit the PR wasn't on my mind anymore, thanks for finding out what I broke ;-) . I reverted the changes and adjusted the comments a bit.

sietseringers:
Great, thanks!

sietseringers (approved):

PR #225, opened 2022-06-23
Update dependencies

sietseringers:
I had to change a helper function in the test code to ensure the Bolt database of `irmaclient.Client` instances is closed after each session test, because otherwise tests fail with the error `too many open files` on macOS.

Background: on macOS, an application may apparently by default only 256 files. An application can override this limit, but doing that in these tests didn't seem appropriate to me. Before upgrading the dependencies, running the integration test opened some 200 files on my machine; afterwards, it reached the 256 limit some after doing about 2/3rd of the tests. By closing the client database after each test we remain well below the limit.


revocation.go
```diff
@@ -766,7 +766,7 @@ func (rs *RevocationStorage) Load(debug bool, dbtype, connstr string, settings R
 		}
 		if err := rs.sqldb.Delete(IssuanceRecord{}, "valid_until < ?", time.Now().UnixNano()); err != nil {
 			err = errors.WrapPrefix(err, "failed to delete expired issuance records", 0)
-			raven.CaptureError(err, nil)
+			sentry.CaptureException(err)
```
ivard, line 769: irmago does not have its own Sentry handling, so these exceptions will be sent into a black hole now. I think we should do something else here.

ivard (approved):

PR #230, opened 2022-07-15
Feat: make postgresdb connection bounds configurable

ivard:
Closes #208

### DIFF
--- a/server/keyshare/keyshareserver/postgresdb.go
+++ b/server/keyshare/keyshareserver/postgresdb.go
@@ -27,7 +27,7 @@ const emailTokenValidity = 24 // amount of time user's email validation token is
 var backoffStart int64 = 60
 
 // newPostgresDB opens a new database connection using the given maximum connection bounds.
-// If maxOpenConns is set to 0, then an unlimited number of connections is allowed.
+// For the maxOpenConns, maxIdleTime and maxOpenTime parameters, the value 0 means unlimited.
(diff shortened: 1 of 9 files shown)


server/keyshare/keyshareserver/postgresdb.go
```diff
@@ -26,11 +26,15 @@ const emailTokenValidity = 24 // amount of time user's email validation token is
...
-func newPostgresDB(connstring string) (DB, error) {
+// newPostgresDB opens a new database connection using the given maximum connection bounds.
+// If maxOpenConns is set to 0, then an unlimited number of connections is allowed.
+func newPostgresDB(connstring string, maxIdleConns int, maxOpenConns int) (DB, error) {
 	db, err := sql.Open("pgx", connstring)
 	if err != nil {
 		return nil, err
 	}
+	db.SetMaxIdleConns(maxIdleConns)
+	db.SetMaxOpenConns(maxOpenConns)
```
ivard, line 37: Open connections will not be closed by the server, because the `ConnMaxLifetime` is unlimited by default. We have to check whether this is a problem. Otherwise, we have to add configuration options for lifetime too.

https://pkg.go.dev/database/sql#DB.SetConnMaxIdleTime

sietseringers: Both of these would be good to have. If you like, I can add them myself to this PR.

ivard:
I added documentation about the default value of newly added parameters of `newPostgresDB`. For the rest, it seems fine by me.

sietseringers (approved):

PR #238, opened 2022-08-25
Fix: Send expiry mails only if deletion is not yet scheduled. Test th…

synaptic-cleft:
…is thoroughly in unit test.

### DIFF
--- a/server/keyshare/tasks/tasks_test.go
+++ b/server/keyshare/tasks/tasks_test.go
@@ -5,7 +5,9 @@ package tasks
 
 import (
 	"database/sql"
+	"fmt"
 	"path/filepath"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


server/keyshare/tasks/tasks_test.go
```diff
@@ -35,6 +37,21 @@ func countRows(t *testing.T, db *sql.DB, table, where string) int {
 	return count
 }
 
+func getDeleteDates(t *testing.T, db *sql.DB) map[string]int {
```
ivard, line 40: In Go, it is uncommon to have a `get` prefix for a getter. This is not a textbook example of a getter of course, but maybe nice to follow that convention here too.

https://go.dev/doc/effective_go#Getters

server/keyshare/tasks/tasks_test.go
```diff
@@ -125,8 +154,19 @@ func TestExpireAccounts(t *testing.T) {
...
+	deleteOnMap := getDeleteDates(t, db)
+
+	assert.Equal(t, 10, countRows(t, db, "users", "delete_on IS NOT NULL"))
+	assert.Equal(t, 4, countRows(t, db, "users", "delete_on IS NULL"))
 
-	assert.Equal(t, 1, countRows(t, db, "users", "delete_on IS NOT NULL"))
+	time.Sleep(1 * time.Second)
+	th.expireAccounts()
+
+	for i, v := range deleteOnMap {
```
ivard, line 165: So `i` is here a unix timestamp right? If I read it I would expect it is an index. That confused me a bit.

server/keyshare/tasks/tasks_test.go
synaptic-cleft: Thanks, old habits die hard. I renamed it to `countDeleteDates` since `deleteDates` would imply deleting dates.

server/keyshare/tasks/tasks_test.go
synaptic-cleft: Changed to `date, count`

ivard (approved):

PR #242, opened 2022-09-12
feat: Custom session nonce in request

stenreijers:
Following the discussion on slack, this feature allows a relying party to set the nonce of the session directly in a signed request.

In a possible scenario:
- A javascript client that talks directly to the IRMA session server (Single-Page Application use-case)
- A relying party that obtains disclosed attributes from the end-user through the Single-Page Application

To prevent replay attacks by the SPA, the relying party must be able to set the nonce in a signed session request, such that the session result cannot be replayed by the javascript client.

### DIFF
--- a/server/irmaserver/sessions.go
+++ b/server/irmaserver/sessions.go
@@ -421,10 +421,15 @@ func (s *Server) newSession(action irma.Action, request irma.RequestorRequest, d
 		request:     request.SessionRequest(),
 	}
 
-	s.conf.Logger.WithFields(logrus.Fields{"session": ses.RequestorToken}).Debug("New session started")
-	nonce, _ := gabi.GenerateNonce()
-	base.Nonce = nonce
+	if base.Nonce == nil {
+		nonce, _ := gabi.GenerateNonce()
+		base.Nonce = nonce
+	}
 	base.Context = one
+
+	// Debug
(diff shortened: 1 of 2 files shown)


stenreijers:
An alternative implementation could be:

We could also choose to leave the `nonce` to be always randomly generated by the IRMA session server. To ensure that the generated proof is unique and no replays can occur: we could introduce a new `challenge` request parameter that should be included in the BaseRequest. The `challenge` can be `nil` or can be set to whatever a relying party wants the challenge to be. 

This ensures that there is always a new proof generated, since the IRMA server will always pick a random `nonce` for each new session it receives.

stenreijers:
You are correct that for normal disclosures there is no nonce included and the JWT could still be replayed. The SessionResult should in that case be updated with a `nonce` field to prevent the replay.

We were using the ABS requests however and in this special case there is a `Signature` attached to the JWT that always includes both the nonce and the full idemix proof list. In this latter case there is no possible way to replay a JWT anymore.

**PROPOSAL**: I would close this pull request in favor of #245 instead. 
**Rationale**: The replay attack is also addressed there with the same file-changes and the new endpoint always provides a `nonce`  (both for normal disclosures and ABS requests).

sietseringers:
Alright, I agree.

PR #243, opened 2022-09-21
Feat: check MX records to verify email on user signup

sanderhollaar:


server/keyshare/email.go
```diff
@@ -129,3 +135,12 @@ func sendHTMLEmail(addr string, a smtp.Auth, from, to *mail.Address, subject str
 		"\r\n")
 	return smtp.SendMail(addr, a, from.Address, []string{to.Address}, append(headers, msg...))
 }
+
+func verifyMXRecord(email string) error {
+	at := strings.LastIndex(email, "@")
+	_, err := net.LookupMX(email[at+1:])
+	if err != nil {
+		return err
+	}
```
ivard, line 144: I'm not sure whether there can be cases that err == nil, but the list of MX records is empty. Did you check that?

Since the MX records list is always initialised, you could also do something like this:
```
records, _ := net.LookupMX(email[at+1:])
if len(records) == 0 {
	return ErrInvalidEmail
}
```
Then we circumvent that issue.

Another little thing, `strings.LastIndex` returns -1 if the string does not contain a `@`. -1+1=0, so in practice your function would crash. But in coding it's nice if you make these things explicit to make clear that you've taken this into account. For example, first do a `if at < 0 { ... }` check.

sanderhollaar: > I'm not sure whether there can be cases that err == nil, but the list of MX records is empty. Did you check that?

verified, a domain witn no MX records causes an error

sanderhollaar: > in coding it's nice if you make these things explicit

done

server/keyshare/email.go
```diff
@@ -129,3 +135,18 @@ func sendHTMLEmail(addr string, a smtp.Auth, from, to *mail.Address, subject str
...
+		if len(records) == 0 {
+			return errors.Errorf("No domain part found in %v", email)
+		} else if err != nil {
+			return err
+		}
+	} else {
+		return errors.Errorf("No '@'-sign found in %v", email)
+	}
+	return nil
+}
```
ivard, line 152: In Go, you often see that the `else` clause is skipped if you use an early `return` in the body of the `if`. That saves indentations. So something like:

```
func verifyMXRecord(email string) error {
	at := strings.LastIndex(email, "@")
        if at < 0 {
            return errors.Errorf("no '@'-sign found in %v", email)
        }
        records, err := net.LookupMX(email[at+1:])
        if err != nil {
            return err
        }
	if len(records) == 0 {
		return errors.Errorf("no domain part found in %v", email)
	}
	return nil
}
```

ivard (changes requested):

server/keyshare/email.go
```diff
@@ -129,3 +135,18 @@ func sendHTMLEmail(addr string, a smtp.Auth, from, to *mail.Address, subject str
...
+	at := strings.LastIndex(email, "@")
+	if at > 0 {
+		records, err := net.LookupMX(email[at+1:])
+		if len(records) == 0 {
+			return errors.Errorf("No domain part found in %v", email)
+		} else if err != nil {
+			return err
+		}
+	} else {
+		return errors.Errorf("No '@'-sign found in %v", email)
```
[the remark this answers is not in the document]
sanderhollaar: Ah yes that makes sense. Thanks for the link!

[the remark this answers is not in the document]
sanderhollaar: Should we also lowercase `22` `var ErrInvalidEmail = errors.New("Invalid email address")` as we return `ErrInvalidEmail` as `error`?

https://github.com/privacybydesign/irmago/blob/feat/verify-email-domain-on-user-signup/server/keyshare/email.go#L22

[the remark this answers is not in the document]
ivard: Well spotted. That should have been the case indeed.

ivard (approved):

ivard (dismissed):

ivard (approved):

PR #246, opened 2022-10-07
Fix keyshare chained sessions bugs

sietseringers:
This fixes two bugs in chained sessions in which a keyshare server is involved.

If in a chain of sessions attributes are disclosed in a session, then they are implicitly (automatically and invisible to the user) always disclosed in all later sessions of the chain. If a later session is an issuance session, then this is thus in fact a combined issuance/disclosure session. This had the following two bugs:

- In combined issuance/disclosure requests, in the array of proofs that the irmaclient sends to the IRMA server, the irmaclient always puts first the `ProofD`'s and then the `ProofU`'s (see `irmaclient.Client.IssuanceProofBuilders()`). However, the keyshare server expected the `ProofU`'s to come first instead of the `ProofD`'s, leading to failures.
- The keyshare code of the irmaclient failed to take into account these implicit disclosures, thus missing that a keyshare server may be involved in these implicit disclosures even when none is in the non-implicit part of the current session. This means that if a session involves disclosure involving a keyshare server while a later session does not involve a keyshare server, then that session fails.

Note: the tests currently fail due to https://github.com/privacybydesign/gabi/pull/37 which will have to be merged first, after which the gabi dependency here will have to be bumped.


ivard:
So, if I understand it correctly, the bug that chained sessions don't work at all is caused by an implementation mistake in the keyshare server (the `ProofD` and `ProofU` order) and by coincidence you found a bug in the `irmaclient` too (when mixing `irma-demo` and `pbdf` schemes within one series of chained sessions).

This means that only deploying this fix to the keyshare server would solve most of the issues. Is my conclusion correct?

ivard:
Probably you already explained this earlier to me and I forgot, but why are those implicit disclosures so important in chained sessions? Naively, I'd say that you could also do completely separate IRMA sessions and chain those together by closing the first session, posting the result to the service provider's backend and wait until it tells how to proceed the second session.

sietseringers:
Since privacybydesign/gabi#37 is now merged, I have updated this PR; it is now ready for review.

> This means that only deploying this fix to the keyshare server would solve most of the issues. Is my conclusion correct?

Yes indeed 🙂 The other issues happen only when `irma-demo` is involved.

> Probably you already explained this earlier to me and I forgot, but why are those implicit disclosures so important in chained sessions? Naively, I'd say that you could also do completely separate IRMA sessions and chain those together by closing the first session, posting the result to the service provider's backend and wait until it tells how to proceed the second session.

To prevent the following attack vector. Suppose the BIG registry makes a chained-sessions flow as follows: you disclose your BSN attribute in session 1, they look up your BIG number if you have one, and if so they issue it in session 2. Without the implicit disclosures two colluding users could do the following:

1. One user discloses her BSN in session 1.
2. In its response to the user's IRMA app, BIG's IRMA server includes a session pointer to session 2. Instead of performing this session herself, the user forwards this session pointer to the other user.
3. That other user receives the BIG attribute instead of the intended user which started the session chain.

To prevent that from happening, all attributes disclosed in session `x` are always implicitly and invisibly are also disclosed in session `x+1`, requiring the exact values disclosed in session `x`. That prevents the attack because the other user won't have the required BSN.

ivard (approved):

PR #253, opened 2022-10-17
Fix/do not abort on invalid email while looping over expired accounts

sanderhollaar:


server/keyshare/tasks/tasks.go
```diff
@@ -83,6 +83,11 @@ func (t *taskHandler) sendExpiryEmails(id int64, username, lang string) error {
 		func(emailRes *sql.Rows) error {
 			var email string
 			err := emailRes.Scan(&email)
+			// FIXME: 'return nil' will prevent abortion of 'expireAccounts()'
+			// but will not take care of the actual problem of handling the invalid email
+			if err == keyshare.ErrInvalidEmail {
+				return nil
+			}
```
ivard, line 90: Now I checked it better, you added the error handling at the wrong place. `emailRes.Scan(&email)` returns errors from the database. That will never give `keyshare.ErrInvalidEmail`. You should have checked this after `t.conf.SendEmail`.

ivard (approved):

ivard:
I give you some wrong advice about where the error check should move to. I'm sorry for that. I corrected my mistake myself and pushed it to this branch.

PR #257, opened 2022-10-19
Middleware to catch a panic and return a 500 error

bobhageman:
@ivard, please don't hold back;)


server/api.go
```diff
@@ -554,6 +554,20 @@ func LogMiddleware(typ string, opts LogOptions) func(next http.Handler) http.Han
 	}
 }
 
+// RecoverMiddware is middleware that logs and returns a 500 error if something unhandled (panic) occurs
+func RecoverMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer func() {
+			err := recover()
+			if err != nil {
```
ivard, line 562: A minor thing, recover does not necessarily return an error, so therefore you often see `r := recover()`.

server/api.go
```diff
@@ -554,6 +554,20 @@ func LogMiddleware(typ string, opts LogOptions) func(next http.Handler) http.Han
...
+				LogError(errors.Errorf("Internal server error: %s", err))
```
ivard, line 563: The `%s` only works for strings and that what recover returns does not necessarily have to be a string I think. So, maybe `%v` would be a bit more safe.

ivard (changes requested):

server/irmaserver/api.go
```diff
@@ -170,6 +170,8 @@ func (s *Server) HandlerFunc() http.HandlerFunc {
 	r := chi.NewRouter()
 	s.router = r
 
+	r.Use(server.RecoverMiddleware)
```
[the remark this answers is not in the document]
bobhageman: Done!

server/api.go
```diff
@@ -554,6 +554,20 @@ func LogMiddleware(typ string, opts LogOptions) func(next http.Handler) http.Han
 	}
 }
 
+// RecoverMiddware is middleware that logs and returns a 500 error if something unhandled (panic) occurs
+func RecoverMiddleware(next http.Handler) http.Handler {
```
ivard, line 558: Small spelling mistake in `RecoverMiddware`.

server/keyshare/myirmaserver/server.go
```diff
@@ -86,6 +86,8 @@ func (s *Server) Handler() http.Handler {
 		router.Use(server.SizeLimitMiddleware)
 		router.Use(server.TimeoutMiddleware(nil, server.WriteTimeout))
 
+		router.Use(server.RecoverMiddleware)
```
ivard, line 89: You've put this everywhere as first middleware to be applied. Maybe good to do that here as well. Then, also a panic in the other middlewares is included.

bobhageman: Very true, and consistency is key, done

ivard (approved):

PR #267, opened 2022-11-14
Implement /getPs endpoint for new keyshare protocol

synaptic-cleft:


ivard (approved):
Looks good. The only thing I noticed was that `c.verifyAccess` is called twice, once in the middleware and once in the request handler. But this is also the case in all the other endpoints, so this is not really an issue of this PR. Since it concerns sensitive information, this might have been intentional. A comment might have helped in that case.

PR #271, opened 2022-11-26
Endpoint versioning

bobhageman:
A requirement to be able to introduce the updated keyshare protocol with newly introduced endpoints. 

Note:
- For the time being the same endpoints are available with and without the `/api/v1` prefix. 
- Inside `server_tests.go` it is now possible to test the endpoints on all api versions using `apiVersions`. This way we can test future versions e.g. `/api/v2` while maintaining backwards compatibility.


synaptic-cleft (approved):

server/keyshare/keyshareserver/server_unversioned_test.go
```diff
@@ -0,0 +1,461 @@
...
+	"github.com/privacybydesign/irmago/internal/test"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+)
+
+func init() {
+	irma.Logger.SetLevel(logrus.FatalLevel)
+}
+
+func TestUnvServerInvalidMessage(t *testing.T) {
```
ivard, line 22: I never saw `Unv` as abbreviation in code before. Is there a big obligation to just write it in full?

irmaclient/keyshare.go
```diff
@@ -156,7 +155,6 @@ func startKeyshareSession(
 		transport := irma.NewHTTPTransport(scheme.KeyshareServer, !ks.client.Preferences.DeveloperMode)
 		transport.SetHeader(kssUsernameHeader, ks.keyshareServer.Username)
 		transport.SetHeader(kssAuthHeader, ks.keyshareServer.token)
-		transport.SetHeader(kssVersionHeader, "2")
```
ivard, line 159: In my opinion we should already add logic in irmaclient that adds the `/api/v1` when it is not mentioned in the URL. If we then release a version of irmago with the keyshare server change, then the irmaclient counterpart is also able to handle both.

server/keyshare/keyshareserver/server_unversioned_test.go
bobhageman: No obligation, think you're right, making the code easier to understand is always a good thing.

irmaclient/keyshare.go
synaptic-cleft: I thought of doing that separately as this does not concern the keyshare server but the interaction between the client and what's stated in the scheme.

However, the reason why I like to keep these things small and separate is for a large extend reviewability. So if you, the reviewer, prefers this in one PR, we can add it. As long as we note that these are actually two unrelated issues.

synaptic-cleft (approved):

irmaclient/keyshare.go
bobhageman: Decided jointly we're not implementing the 'replace url' logic in irmaclient this step. It will be handled separate and differently.

ivard (approved):

PR #273, opened 2022-12-02
Get commitments

synaptic-cleft:


irmaclient/session.go
```diff
@@ -488,20 +489,26 @@ func (session *session) doSession(proceed bool, choice *irma.DisclosureChoice) {
...
+			session.Version,
 		)
+		if !auth {
+			// newKeyshareSession() calls session.fail() in case of failure, no need to do that here
+			return
+		}
+		session.builders, session.attrIndices, session.issuerProofNonce, err = session.getBuilders(keyshareSession)
+		if err != nil {
+			session.fail(&irma.SessionError{ErrorType: irma.ErrorCrypto, Err: err})
+		}
```
ivard, line 507: Shoudn't we return after the `session.fail`?

messages.go
```diff
@@ -391,6 +392,53 @@ func (ppcm *ProofPCommitmentMap) MarshalJSON() ([]byte, error) {
...
+}
+
+type ProofPCommitmentMapV2 struct {
+	Commitments map[PublicKeyIdentifier]*big.Int `json:"c"`
+}
+
+func (cm *ProofPCommitmentMapV2) MarshalJSON() ([]byte, error) {
+	var encCM struct {
+		Commitments map[string]*big.Int `json:"p"`
+	}
```
ivard, line 428: Shouldn't the `Commitments` field have the JSON tag `c`?

server/keyshare/keyshareserver/server.go
```diff
@@ -122,6 +122,14 @@ func (s *Server) Handler() http.Handler {
...
 			s.routeHandler(r)
 		})
+
+		router.Route("/api/v2", func(r chi.Router) {
+			// Keyshare sessions with provably secure keyshare protocol
+			r.Use(s.userMiddleware)
+			r.Use(s.authorizationMiddleware)
+			r.Post("/prove/getPs", s.handlePs)
+			r.Post("/prove/getCommitments", s.handleCommitmentsV2)
+		})
```
ivard, line 132: So adding the `getResponse` here is still a TODO?

server/keyshare/keyshareserver/server.go
```diff
@@ -178,6 +186,56 @@ func (s *Server) loadIdemixKeys(conf *irma.Configuration) error {
...
+
+	// Read keys
+	var keys []irma.PublicKeyIdentifier
+	if err := server.ParseBody(r, &keys); err != nil {
+		server.WriteError(w, server.ErrorInvalidRequest, err.Error())
+		return
+	}
+	if len(keys) == 0 {
+		s.conf.Logger.Info("Malformed request: no keys for P specified")
+		server.WriteError(w, server.ErrorInvalidRequest, "No key specified")
```
ivard, line 203: In Go, error messages like `No key specified` are expected in lowercase. This suits better the way how errors are printed in logging. I know we do not do this consistently, but maybe a good practice to not make it worse.

https://earthly.dev/blog/golang-errors/

server/keyshare/keyshareserver/server.go
```diff
@@ -240,6 +298,71 @@ func (s *Server) generateCommitments(user *User, authorization string, keys []ir
...
+
+	// Read keys
+	var req irma.GetCommitmentsRequest
+	if err := server.ParseBody(r, &req); err != nil {
+		server.WriteError(w, server.ErrorInvalidRequest, err.Error())
+		return
+	}
+	if len(req.Keys) == 0 {
+		s.conf.Logger.Info("Malformed request: no keys for commitment specified")
+		server.WriteError(w, server.ErrorInvalidRequest, "No key specified")
```
ivard, line 315: Here too.

server/keyshare/keyshareserver/server.go
```diff
@@ -240,6 +298,71 @@ func (s *Server) generateCommitments(user *User, authorization string, keys []ir
...
+		return
+	}
+
+	commitments, err := s.generateCommitmentsV2(user, authorization, req)
+	// TODO: can ErrInvalidChallenge be removed?
+	if err != nil && (err == keysharecore.ErrInvalidChallenge || err == keysharecore.ErrInvalidJWT) {
+		server.WriteError(w, server.ErrorInvalidRequest, err.Error())
+		return
+	}
```
ivard, line 324: Is this something you want to sort out in a later PR?

server/keyshare/keyshareserver/server.go
```diff
@@ -240,6 +298,71 @@ func (s *Server) generateCommitments(user *User, authorization string, keys []ir
...
+		s.conf.Logger.WithField("error", err).Warn("Could not generate commitments for request")
+		return nil, err
+	}
+
+	// Prepare output message format
+	// TODO: move logic to gabi?
+	mappedCommitments := map[irma.PublicKeyIdentifier]*big.Int{}
+	for i, keyID := range req.Keys {
+		mappedCommitments[keyID] = commitments[i].Pcommit
+	}
```
ivard, line 347: What's the aim of this TODO?

ivard (changes requested):

internal/keysharecore/operations.go
```diff
@@ -264,7 +295,7 @@ func (c *Core) GenerateResponse(secrets UserSecrets, accessToken string, commitI
 
 	// Generate response
 	token := jwt.NewWithClaims(jwt.SigningMethodRS256, jwt.MapClaims{
-		"ProofP": gabi.KeyshareResponse(s.KeyshareSecret, commit, challenge, key),
+		"ProofP": gabi.KeyshareResponseLegacy(s.KeyshareSecret, commit, challenge, key),
```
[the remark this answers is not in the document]
synaptic-cleft: The new keyshare protocol endpoint (i.e. v2) for getResponse is not implemented yet. You're looking at the old endpoint here which is why  `KeyshareResponseLegacy` is used here

messages.go
synaptic-cleft: yep, seems a copy-paste mistake. i'll fix it

server/keyshare/keyshareserver/server.go
synaptic-cleft: yep, i've hardly added any todo's since we try to make the PRs small. the story mapping gives an idea of what is yet to be done.

server/keyshare/keyshareserver/server.go
synaptic-cleft: yep

server/keyshare/keyshareserver/server.go
synaptic-cleft: Due to legacy we get more back from gabi than we need and need to do an extra mapping here. I put down a TODO because I'd like to evaluate again later when the rest is implemented if we still want to move some things from gabi to irmago or the other way round.

In this case gabi could also return a `map[irma.PublicKeyIdentifier]*big.Int{}` directly.

irmaclient/keyshare.go
```diff
@@ -81,29 +82,27 @@ func newKeyshareServer(schemeManagerIdentifier irma.SchemeManagerIdentifier) (*k
...
 	client *Client,
 	pin KeysharePinRequestor,
-	builders gabi.ProofBuilderList,
 	session irma.SessionRequest,
 	implicitDisclosure [][]*irma.AttributeIdentifier,
-	issuerProofNonce *big.Int,
-	timestamp *atum.Timestamp,
-) {
+	protocolVersion *irma.ProtocolVersion,
+) (*keyshareSession, bool) {
```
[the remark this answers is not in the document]
sietseringers: It's still passed to the `keyshareSession` struct, only not here anymore but a few lines below. The reason for this has not changed: in case of attribute-based signatures sessions, the timestamp that is to be included in the attribute-based signature is an input parameter for the challenge when it is constructed as the hash over several parameters, [here](https://github.com/privacybydesign/irmago/blob/f2274128d6c5d0dd772f6a5180e44b22a1a681b2/irmaclient/keyshare.go#L422). This is part of the cryptographic mechanism that convinces the verifier that all disclosures and signatures that it receives come from a single wallet within a single session.

The reason that this timestamp was first passed as parameter to `startKeyshareSession` and now is set some lines below is because the order of some things has had to change a little:
* To construct the proof builders at the start of a session, I now need some keyshare session state in order to be able to contact the new endpoint `/getPs` in case of issuance sessions. That is why the `keyshareSession`, i.e. the output of `startKeyshareSession()`/`newKeyshareSession()`, is now passed to `session.getBuilders()`.
* Only after `session.getBuilders()` has completed has the timestamp been constructed, so only then can we put it on the `keyshareSession`.

irmaclient/session.go
sietseringers: Yes! Thanks, added.

ivard (approved):

PR #276, opened 2023-01-12
Fix: side effects of scheme update mechanism when an error occurs

ivard:
Issues like https://github.com/privacybydesign/irmamobile/issues/66 

`checkRemoteScheme` should not edit the index file (the `writeIndex` function), because then the active scheme is edited. This means that if the function is interrupted after this action, the scheme gets in an inconsistent state.

`InstallScheme` has a similar issue, but that was not so easy to solve. This function is not used that often; only on initial start-up and in recovery situations. Therefore, the issue is not so critical here and I choose to only implemented undo functionality in case an error occurs. I documented this limitation.


irmago_test.go
```diff
@@ -1550,3 +1551,53 @@ func TestParseKeysFolderConcurrency(t *testing.T) {
...
+		require.NoError(t, unstableSchemeServer.Close())
+	}()
+
+	// Initialize empty configuration
+	conf, err := NewConfiguration(t.TempDir(), ConfigurationOptions{})
+	require.NoError(t, err)
+
+	// Check whether installing fails cleanly when using the unstable remote
+	pkPath := path.Join(corruptTestData, "irma_configuration", "test", "pk.pem")
+	pkBytes, err := os.ReadFile(path.Join(corruptTestData, "irma_configuration", "test", "pk.pem"))
```
ivard, line 1579: Use `pkPath` variabele here

bobhageman (approved):

sjoerdlmkns (approved):

PR #278, opened 2023-01-23
Keyshare server getResponse v2 endpoints

synaptic-cleft:


internal/keysharecore/operations.go
```diff
@@ -304,6 +304,62 @@ func (c *Core) GenerateResponse(secrets UserSecrets, accessToken string, commitI
 	return token.SignedString(c.jwtPrivateKey)
 }
 
+// GenerateResponseV2 generates the response of a zero-knowledge proof of the keyshare secret, for a given previous commit and response request.
+func (c *Core) GenerateResponseV2(
+	secrets UserSecrets,
+	accessToken string,
+	commitID uint64,
+	hash gabi.KeyshareCommitmentRequest,
```
ivard, line 312: Maybe you can make the naming more clear here. I first thought that this was just a string or `[]byte` or something, but it appears to be a struct with a list of hashes.

internal/keysharecore/operations.go
```diff
@@ -304,6 +304,62 @@ func (c *Core) GenerateResponse(secrets UserSecrets, accessToken string, commitI
...
+	}
+
+	if uint(proofP.C.BitLen()) > gabikeys.DefaultSystemParameters[1024].Lh || proofP.C.Cmp(big.NewInt(0)) < 0 {
+		return "", ErrInvalidChallenge
+	}
+
+	// Linkable response for legacy purposes
+	if linkable {
+		proofP.P = new(big.Int).Exp(key.R[0], s.KeyshareSecret, key.N)
+	}
```
ivard, line 350: Is this `linkable` part documented somewhere? It looks like magic to me what it does.

ivard: @sietseringers Do you have some documentation why this is needed? Then I can add the comment myself.

sietseringers: Sorry for the late response! I missed this question completely.

The reason for this is as follows. In both versions of the keyshare protocol, the user sends a `ProofP` instance to the issuer to prove to it that it knows its part of the full secret key (which is, in both versions of the protocol, the sum of the user's part and of the keyshare server's part). In the old protocol, this included the value `P` computed here. Schematically this is computed as `P = R_0^userSecret`, where `R_0` comes from the issuer public key and where `userSecret` is the user's part of the secret key. Since both of those two ingredients are fixed (i.e. not ephemeral and random per session), this has the consequence that in the issuance protocol (contrary to the disclosure protocol), the user is always linkable across issuance sessions: if you do a second issuance session with one issuer, then the issuer can theoretically recognize you from the first session because you have to send the exact same `P` both times. That is what this parameter name `linkable` refers to. Since the new protocol is such that this value `P` is not required, this is not an issue in the new issuance protocol (which is one of its advantages).

However, the client might talk with an old issuer that only understands the old issuance protocol, and so we need to be backwards compatible. In such cases we need to include this `P` value in the `ProofP`. That is what this piece of code does. (Actually as I recall there should be other modifications elsewhere as well, but that is not relevant to this question.) The name `linkable` for this parameter is not terribly clear to my taste without all this context; perhaps it can better be called `legacy` or `legacyProtocol` or something.

ivard: Thanks for the information. I added a brief explanation in the code.

ivard (approved):

PR #280, opened 2023-01-25
chore: make pin- and email-settings configurable

sanderhollaar:


server/keyshare/keyshareserver/postgresdb.go
```diff
@@ -19,10 +19,7 @@ type postgresDB struct {
 	db keyshare.DB
 }
 
-const maxPinTries = 3                  // number of tries allowed on pin before we start with exponential backoff
-const emailTokenValidity = 24          // amount of time user's email validation token is valid (in hours)
-const emailTokenRateLimit = 3          // max number of active tokens per email address within the emailTokenRateLimitDuration
-const emailTokenRateLimitDuration = 60 // amount of time after which tokens become irrelevant for rate limiting (in minutes)
+var conf *Configuration
```
ivard, line 22: Mostly, you don't want to work with global variables if not necessary. This leads to all kinds of ugly situations in unit and integration tests if you'd run tests in parallel. Both tests will read and write to the same global variable, which can lead to race conditions. It's best practice to add this configuration to the state struct, in this case the `type postgresDB struct`.

server/keyshare/keyshareserver/conf.go
```diff
@@ -63,6 +63,14 @@ type Configuration struct {
 	registrationEmailTemplates map[string]*template.Template
 
 	VerificationURL map[string]string `json:"verification_url" mapstructure:"verification_url"`
+	// number of tries allowed on pin before we start with exponential backoff
+	MaxPinTries int `json:"max_pin_tries" mapstructure:"max_pin_tries"`
```
ivard, line 67: This is a chore security parameter, and I doubt whether we want to make this configurable. There is not really another logical value than the one we had hardcoded in my view. This means that making this setting configurable only introduces a risk on misconfiguration. I'd keep this one constant.

server/keyshare/keyshareserver/conf.go
```diff
@@ -63,6 +63,14 @@ type Configuration struct {
 	registrationEmailTemplates map[string]*template.Template
 
 	VerificationURL map[string]string `json:"verification_url" mapstructure:"verification_url"`
+	// number of tries allowed on pin before we start with exponential backoff
```
ivard, line 66: A small nit pick: for all the other fields in this struct, the comment starts with a capital letter. If you're editing anyway, it's maybe nice to make this consistent by starting with a capital everywhere in this struct.

server/keyshare/keyshareserver/conf.go
```diff
@@ -63,6 +63,14 @@ type Configuration struct {
...
+	MaxPinTries int `json:"max_pin_tries" mapstructure:"max_pin_tries"`
+	// amount of time user's email validation token is valid (in hours)
+	EmailTokenValidity int `json:"email_token_validity" mapstructure:"email_token_validity"`
+	// max number of active tokens per email address within the emailTokenRateLimitDuration
+	EmailTokenRateLimit int `json:"email_token_rate_limit" mapstructure:"email_token_rate_limit"`
+	// amount of time after which tokens become irrelevant for rate limiting (in minutes)
+	EmailTokenRateLimitDuration int `json:"email_token_rate_limit_duration" mapstructure:"email_token_rate_limit_duration"`
```
ivard, line 73: Probably you also want to add these options to the `irma` CLI tool. Otherwise, you won't be able to actually modify these configuration options. It's also nice to specify some default values there. If no value is explicitly set (using a command line parameter or in the `--config` file), then there is a fallback.
https://github.com/privacybydesign/irmago/blob/master/irma/cmd/keyshare-server.go#L32

You probably also want to specify some test settings for these options. Otherwise the default value for `int` will be used, which is 0. 0 is not a logical value for most of these new configuration options, so that will probably cause some tests to fail.
https://github.com/privacybydesign/irmago/blob/master/internal/testkeyshare/testkeyshare.go#L60

ivard (changes requested):

server/keyshare/keyshareserver/postgresdb.go
```diff
@@ -217,7 +222,7 @@ func (db *postgresDB) addEmailVerification(user *User, emailAddress, token strin
 	if err != nil {
 		return err
 	}
-	if amount >= emailTokenRateLimit {
+	if int(amount) >= emailTokenRateLimit {
```
ivard, line 225: `amount` already is an `int`, so you don't need to explicitly convert this.

server/keyshare/keyshareserver/postgresdb.go
```diff
@@ -206,9 +211,9 @@ func (db *postgresDB) addLog(user *User, eventType eventType, param interface{})
 	return err
 }
 
-func (db *postgresDB) addEmailVerification(user *User, emailAddress, token string) error {
-	expiry := time.Now().Add(emailTokenValidity * time.Hour)
-	maxPrevExpiry := expiry.Add(-1 * emailTokenRateLimitDuration * time.Minute)
+func (db *postgresDB) addEmailVerification(user *User, emailAddress, token string, validity int) error {
+	expiry := time.Now().Add(time.Duration(validity) * time.Hour)
+	maxPrevExpiry := expiry.Add(-1 * time.Duration(emailTokenRateLimitDuration) * time.Minute)
```
ivard, line 216: More like a fun fact if you were wondering what was happening here:
In `time.Duration(emailTokenRateLimitDuration)` it is not really required to add the `time.Duration(...)` part.
Something like `maxPrevExpiry := expiry.Add(-1 * emailTokenRateLimitDuration * time.Minute)` would work too, because `time.Duration` itself is also represented as an `int`. This is the reason why it worked in the old situation.

ivard:
By the way, I assume the `WIP` prefix can be removed from the PR now?

ivard (approved):

sjoerdlmkns (approved):

PR #283, opened 2023-02-10
Feat: make it possible to increase MaxSessionLifetime without increasing the SessionResult lifetime

ivard:
Changed default to 15 minutes, while keeping the SessionResult lifetime to 5 minutes.

### DIFF
--- a/irma/cmd/server.go
+++ b/irma/cmd/server.go
@@ -115,7 +115,8 @@ func setFlags(cmd *cobra.Command, production bool) error {
 	flags.StringSlice("revoke-perms", nil, "list of credentials that all requestors may revoke")
 	flags.Bool("skip-private-keys-check", false, "whether or not to skip checking whether the private keys that requestors have permission for using are present in the configuration")
 	flags.String("static-sessions", "", "preconfigured static sessions (in JSON)")
-	flags.Int("max-session-lifetime", 5, "maximum duration of a session once a client connects in minutes")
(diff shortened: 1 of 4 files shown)


irma/cmd/server.go
```diff
@@ -115,7 +115,8 @@ func setFlags(cmd *cobra.Command, production bool) error {
 	flags.StringSlice("revoke-perms", nil, "list of credentials that all requestors may revoke")
 	flags.Bool("skip-private-keys-check", false, "whether or not to skip checking whether the private keys that requestors have permission for using are present in the configuration")
 	flags.String("static-sessions", "", "preconfigured static sessions (in JSON)")
-	flags.Int("max-session-lifetime", 5, "maximum duration of a session once a client connects in minutes")
+	flags.Int("max-session-lifetime", 15, "maximum duration of a session once a client connects in minutes")
```
ivard, line 118: This means the default value of `ClientTimeout` also increases to 15 minutes. I'll check whether that's okay.

bobhageman: @ivard, to be sure: has it been found ok?

ivard: Yes, @MartijnYivi did not see any blocking issues. But if you have a strong opinion about this, then feel free to say so.

bobhageman: Nope, think it is just fine, just double checking!;)

bobhageman (approved):

sjoerdlmkns (approved):

sanderhollaar (approved):

PR #294, opened 2023-03-13
Test: added performance test script for irma server

ivard:


testdata/performance/keyshare-server.js
```diff
@@ -41,8 +49,9 @@ export function setup() {
         'X-IRMA-MaxProtocolVersion': '2.8',
       },
     });
+    checkResponse(sessionResp);
```
ivard, line 52: Call `checkResponse` on registerResp too.

testdata/performance/irma-server.js
```diff
@@ -0,0 +1,69 @@
...
+  thresholds: {
+    http_req_failed: ['rate<0.01'], // http errors should be less than 1%
+    http_req_duration: ['p(95)<200'], // 95% of requests should be below 200ms
+  },
+};
+
+function checkResponse(response, expectedOutput = '') {
+  const checkOutput = check(response, {
+    'verify status code': (r) => r.status === 200,
+    'verify body': (r) => r.body.includes(expectedOutput),
```
ivard, line 17: Body can be `null`. This should be taken into account.

testdata/performance/keyshare-server.js
```diff
@@ -14,6 +14,14 @@ export const options = {
   },
 };
 
+function checkResponse(response, expectedOutput = '') {
+  const checkOutput = check(response, {
+    'verify status code': (r) => r.status === 200,
+    'verify body': (r) => r.body.includes(expectedOutput),
```
ivard, line 20: Body can be `null`. This should be taken into account.

PR #297, opened 2023-03-24
add mssql (Microsoft SQL Server) as supported db type for revocation

ams-azaaiman:
Besides postgres and mysql, support for mssql is preferrable for our environment.
Mostly to prevent yet another database type to be added to our hosting environment.

Support has only been added for the revocation part, since that is the part we need.

Due to a case-sensitivity bug in the GORM mssql package, there was a need to uppercase some of the GORM instructions.

The [README.md](README.md) has been updated with instructions to run the tests against SQL Server.

We have been running with this configuration for a few months in our staging environment, without any problems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- Support for revocation db type `mssql` (Microsoft SQL Server)
+
 ## [0.12.2] - 2023-03-22
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


ivard:
Thanks for your nice contribution. Do you need this change to be included at a certain date? Then we take that into account.

Sidenote, it would be nice if you can also add public documentation for this new feature.
https://irma.app/docs/revocation/#revocation-settings
https://github.com/privacybydesign/irma-documentation

ams-azaaiman:
> Do you need this change to be included at a certain date? Then we take that into account.

Not really, we build our own executable.

ivard:
I lately improved the stability and solved some bugs in the revocation database logic. Therefore, the merge conflict became quite big. In #324 I tried to resolve these. I also made sure the revocation unit tests are run with all database types now.

One of the stability improvements was that I did a major upgrade of GORM, and the MS SQL Server driver in the new GORM version works a bit differently. For example, it is named `sqlserver` instead of `mssql`. Therefore, I also use the `sqlserver` identifier to maintain the consistency. That's however a change compared to your proposal in this PR.

Can you compare your PR with my proposal in #324 and give me some feedback?

PR #300, opened 2023-05-01
Fix: keyshare server returns 403 status codes when database is down

ivard:


server/keyshare/myirmaserver/postgresdb.go
```diff
@@ -149,11 +149,14 @@ func (db *postgresDB) loginUserCandidates(token string) ([]loginCandidate, error
 
 func (db *postgresDB) verifyLoginToken(token, username string) (int64, error) {
```
bobhageman, line 150: Note: inside method `processTokenLogin` in `server/keyshare/myirmaserver/server.go:298` the error `keyshare.ErrUserNotFound` can never happen since `QueryUser` is changed into `QueryScan`.

ivard: Good catch. There was a minor inconsistency between the memorydb and the postgresdb implementation, as a result of which the unit tests did not detect this issue. I fixed it.

bobhageman (approved):

sjoerdlmkns (approved):

PR #313, opened 2023-06-02
Feat: Email revalidation

bobhageman:
This feature fixes the issues with expiring accounts having one or more (temporarily) non-working e-mail addresses causing possible traffic jams during task processing.


server/keyshare/tasks/tasks.go
```diff
@@ -115,40 +140,65 @@ func (t *taskHandler) sendExpiryEmails(ctx context.Context, id int64, username,
...
 
-// Mark old unused accounts for deletion, and inform their owners.
+// Mark old unused accounts for deletion, and inform their owners. When email revalidation enabled,
+// email addresses which where marked for revalidation are skipped because these will be processed
+// separately inside revalidateEmails.
 func (t *taskHandler) expireAccounts(ctx context.Context) {
 	// Disable this task when email server is not given
 	if t.conf.EmailServer == "" {
-		t.conf.Logger.Warning("Expiring accounts is disabled, as no email server is configured")
+		t.conf.Logger.Warning("expiring accounts is disabled, as no email server is configured")
```
ivard, line 174: Logging may be capitalised. The requirement that errors should be lowercase only holds for go errors.

ivard (changes requested):

server/keyshare/tasks/tasks.go
```diff
@@ -115,40 +140,65 @@ func (t *taskHandler) sendExpiryEmails(ctx context.Context, id int64, username,
...
+				// When email revalidation is enabled and sending was impossible because of
+				// (temporary) MX / A record issues at the domain or an invalid email address,
+				// we mark the record to be revalidated in 5 days from now.
+				if err = t.db.ExecUserContext(ctx, "UPDATE irma.emails SET revalidate_on = $1 WHERE id = $2",
+					time.Now().AddDate(0, 0, 5).Unix(),
+					id); err != nil {
+					t.conf.Logger.WithField("error", err).Error("could not update email address to set revalidate_on")
+					return err
+				}
+				return nil
```
ivard, line 154: If I'm right the situation might occur now that if a user has two email addresses of which the first is valid and the second one isn't, then the user will already receive mail on the first email address that the account is going to expire while the account is actually flagged for revalidation due to the second email address.

server/keyshare/tasks/tasks.go
```diff
@@ -66,6 +75,19 @@ func runWithTimeout(fn func(ctx context.Context)) error {
 	return ctx.Err()
 }
 
+func hasMailRevalidation(conf *Configuration, db *keyshare.DB) bool {
+	ctx, cancel := context.WithTimeout(context.Background(), taskTimeout)
+	defer cancel()
+
+	c, err := db.ExecCountContext(ctx, "SELECT true FROM information_schema.columns where table_schema='irma' AND table_name='emails' AND column_name='revalidate_on'")
+	if err != nil {
+		conf.Logger.WithField("error", err).Error("could not query the schema for column emails.revalidate_on, therefore revalidation is disabled")
```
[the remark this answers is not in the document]
bobhageman: If this error occurs it is because querying itself actually failed. This could happen due to a db server issue, a rights issue on the information_schema.columns table etc. Normally this won't happen: if the `revalidate_on` column does not exist, c == 0 and the function just returns false. 

But if the query does throw an error, I could be a severe issue, so we'd better log it as an error. Because it might be an isolated problem we do not halt but just return false, allowing further processing of the tasks. 

To prevent duplicate logging I've moved the warning log into `hasMailRevalidation` so it now either returns true, or false (with an error or a warning log entry).

server/keyshare/tasks/tasks.go
```diff
@@ -159,28 +209,85 @@ func (t *taskHandler) expireAccounts(ctx context.Context) {
...
+				return err
+			}
+
+			addr, err := keyshare.ParseEmailAddress(email)
+
+			if err != nil {
+				return err
+			}
+
+			if err := keyshare.VerifyMXRecord(addr.Host); err != nil {
```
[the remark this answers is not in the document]
bobhageman: Would be nice, but don't think that'll work in here because validation of the email address should be done while constructing the object with the given email address string as input. So I do not see a way to do this without adding complexity. Do you? If so: please enlighten me:)

This is what I come up with:
```
type EmailAddress struct {
	Address string
	Host    string
}

func (e EmailAddress) Validate() error {
	addr, err := mail.ParseAddress(t.Address)
	if err != nil {
		return ErrInvalidEmail
	}

	e.Address = addr.Address
	e.Host = addr.Address[strings.LastIndex(addr.Address, "@")+1:]

	return nil
}

// and then:
	addr := EmailAddress{ Address: "email@domain.com" }
	
	if !addr.Validate() {
		...
	}
```

But this allows inconsistencies at Address and Host (can be overwritten/altered) and Validate (which sets the Host) is not forced so could be forgotten.

Also wanted to keep it's behavior the same as mail.ParseAddress(addr)

What do you think?

server/keyshare/email.go
```diff
@@ -18,6 +18,10 @@ type EmailConfiguration struct {
 	DefaultLanguage string `json:"default_language" mapstructure:"default_language"`
 	EmailAuth       smtp.Auth
 }
+type EmailAddress struct {
+	mail.Address
```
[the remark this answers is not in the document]
bobhageman: Of course! Thanks for the insight, applied.

server/keyshare/tasks/tasks.go
bobhageman: You found something here, made a small tweak in sendExpiryEmails so a scenario where the user could receive multiple emails on the same address cannot happen anymore. Let me put it in writing to clarify:

Lets say:
User A has x (1, 2 or more) valid email addresses.
User B has 1 valid and 1 invalid email address.
User C has 1 invalid email address.
All users have more than 1 year of inactivity.

Now expireAccounts runs.

Without revalidation: 
- User A gets x expiry emails, account is flagged for deletion in 30d.
- User B gets 1 expiry email, the user 'hangs' forever inside expireAccounts making the effective LIMIT -=1.
- User C gets no expiry mail, the user 'hangs' forever inside expireAccounts making the effective LIMIT -=1.

With revalidation:
- User A gets x expiry mails, account is flagged for deletion in 30d.
- User B gets no expiry mail (yet). The invalid email address is flagged for revalidation. If the email address is still invalid after 5 days, it is deleted. If the email address is valid after 5 days, revalidate_on is removed from it. In both ways: with next run of expireAccounts now the user gets 1 or 2 expiry mails and account is flagged for deletion in 30d. 
- User C gets no expiry mail (yet). If the email address is still invalid after 5 days, it is deleted. The user now has no email addresses and stays (for now) forever in the db. If the email address is valid after 5 days, revalidate_on is removed and with the next run of expireAccounts the user receives 1 expiry mail and account is flagged for deletion in 30d.

CHANGELOG.md
```diff
@@ -6,8 +6,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- E-mail address revalidation. At automated mailings (`irma keyshare tasks`) informing users their account is about to expire (temporary) invalid e-mail addresses are withheld from processing for 5 days, preventing cluttering of the processing queue
+
+**Note:** E-mail address revalidation requires a change in the database schema. In order to enable this feature please add the `revalidate_on` column of type `bigint` to the `irma.emails` table. See the [schema](https://github.com/privacybydesign/irmago/tree/master/server/keyshare/schema.sql#L50) file.
```
ivard, line 12: Maybe you can add that the current code is not breaking. It keeps working with the old database structure.

CHANGELOG.md
```diff
@@ -6,8 +6,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 ### Changed
 - Use separate application user in Dockerfile for entrypoint
+As part of e-mail address revalidation:
+- `VerifyMXRecord` incorporates a check to see if there is an active network connection
+- MyIrma server: `/user` returns an additional field `revalidate_in_progress` in the JSON response body, indicating whether the e-mail address is being revalidated or not 
+- MyIrma server: `/user/delete` and `/email/remove` return a 400 status code if one or more e-mail addresses of the user are invalid
+- MyIrma server: `/email/remove` returns a 400 status code if one or more e-mail addresses of the user are invalid
```
ivard, line 20: It's not really something the user can change at this point anymore. Wouldn't a 5xx error make more sense in this case then?

server/keyshare/myirmaserver/server.go
```diff
@@ -162,16 +162,41 @@ func (s *Server) sendDeleteEmails(ctx context.Context, session *session) error {
...
+				return err
+			}
+			if err := s.db.setPinBlockDate(ctx, *session.userID, delay); err != nil {
+				s.conf.Logger.WithField("error", err).Error("Could not update pin block date")
+				return err
+			}
+
+			s.conf.Logger.WithError(err).Warn("Could not send delete email")
+			return err
+		}
```
ivard, line 199: Suppose the user has three email addresses, and the third email address is invalid. It looks to me that we already sent two emails that the account will be deleted, and then that behaviour is aborted when sending the third email fails. In my view as much validation logic as possible should be executed for all email addresses before we start sending emails, such that we prevent mixed messages in emails.

server/keyshare/tasks/tasks.go
```diff
@@ -66,6 +76,24 @@ func runWithTimeout(fn func(ctx context.Context)) error {
...
+		return false
+	}
+
+	if c == 0 {
+		conf.Logger.Warning("Email address revalidation is disabled because the emails.revalidate_on column is not present in the schema")
+		return false
+	}
+
+	return true
+}
```
ivard, line 95: It looks like `myirmaserver` has a similar function. You could consider moving this to being a method of the `keyshare.DB` struct, such that both `myirmaserver` and `tasks` can use this implementation.

server/keyshare/tasks/tasks.go
```diff
@@ -115,40 +140,65 @@ func (t *taskHandler) sendExpiryEmails(ctx context.Context, id int64, username,
 				email,
 				lang,
 			)
-			return err
+
+			if t.revalidateMail && err != nil && err != keyshare.ErrNoNetwork {
+				// When email revalidation is enabled and sending was impossible because of
+				// (temporary) MX / A record issues at the domain or an invalid email address,
+				// we mark the record to be revalidated in 5 days from now.
+				if err = t.db.ExecUserContext(ctx, "UPDATE irma.emails SET revalidate_on = $1 WHERE id = $2",
```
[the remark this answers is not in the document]
ivard: Ah I mixed up the `id` and `user_id` column in the database. That explains my confusion.

server/keyshare/tasks/tasks.go
```diff
@@ -126,29 +176,40 @@ func (t *taskHandler) sendExpiryEmails(ctx context.Context, id int64, username,
 	return nil
 }
 
-// Mark old unused accounts for deletion, and inform their owners.
+// expireAccounts marks old unused accounts for deletion and informs their owners.
+// When email revalidation is enabled, email addresses which where marked for revalidation are skipped
```
ivard, line 180: `which were marked`

server/keyshare/tasks/tasks.go
```diff
@@ -126,29 +176,40 @@ func (t *taskHandler) sendExpiryEmails(ctx context.Context, id int64, username,
...
 		SELECT id, username, language
 		FROM irma.users
 		WHERE last_seen < $1 AND (
 			SELECT count(*)
 			FROM irma.emails
 			WHERE irma.users.id = irma.emails.user_id
+			{{revalidate}}
 		) > 0 AND delete_on IS NULL
-		LIMIT 10`,
+		LIMIT 10`
```
ivard, line 198: This means that if the user has two email addresses registered, and one of those addresses is being revalidated, then we continue to expire its account... I'd say that we wait until we know what the result of the revalidation will be. This is only a few days.

server/keyshare/tasks/tasks.go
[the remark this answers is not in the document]
ivard: It's not so practical to make changes inside `Validate` indeed. Then you'd need to parse the email address twice, once to create the struct and once to validate it. That might be a bit overkill indeed.

ivard (changes requested):

server/keyshare/tasks/tasks.go
ivard: Sounds great. This is not implemented yet, right?

You should also check all derived cases of this issue type, like
https://github.com/privacybydesign/irmago/pull/313#discussion_r1241968456

bobhageman: It is implemented, with the following addition now: on all relevant places first validation logic is applied to all users' email addresses. And then only the valid one's receive a mail (in case of multiple it's a BCC mail).

server/keyshare/myirmaserver/server.go
bobhageman: Changed it in such a way that for each user first all it's email addresses are validated. Then only the valid ones receive a single email (in case of multiple it becomes a BCC mail) and the invalid ones will be processed accordingly.

server/keyshare/tasks/tasks.go
bobhageman: Applied the following logic:
- If one of multiple email addresses is valid, the user will receive an email about expiry. In that case I do think it is logical if we continue expiring the account, so that's what happens.
- When none of multiple addresses is valid, then account expiry will not happen until revalidation has done it's thing.

CHANGELOG.md
bobhageman: I could not find a fitting 5xx status code. If we'd use the general 500 status code we're not able to distinguish between a internal server error and invalid email addresses.

You can see the 400 error behavior on irma_keyshare_webclient [here](https://github.com/privacybydesign/irma_keyshare_webclient/pull/37/files#diff-68cfdabf9fa2ce99fa19fdbe25785f2622056c22f88b1f846d3ac7ffffea0744R69) and [here](https://github.com/privacybydesign/irma_keyshare_webclient/pull/37/files#diff-68cfdabf9fa2ce99fa19fdbe25785f2622056c22f88b1f846d3ac7ffffea0744R93). 

It's not ideal I agree. Otherwise we could maybe go for a 422 Unprocessable Content?

ivard: Yeah it's not ideal. Let's keep the 400 then.

server/keyshare/tasks/tasks.go
ivard: I've a bit of a dilemma. If you registered multiple email addresses, would you either want to get a message on all your email addresses or would you think its okay if you only receive it on a subset of the email address. In case it was caused by an explicit action of the user, then the latter one seems more logical to me. Otherwise, if one email address has issues you cannot do the action. But if it concerns something automatic, like account expiry, I tend to prefer the former one, but it's only a subtle difference. What do you think?

bobhageman: Updated the tasks implementation to match the discussed and agreed solution: only if all email addresses are valid the expiry mail is sent. This means if a user has one or more invalid addresses, these will first go through the revalidation process.

server/keyshare/tasks/tasks.go
ivard: This code is still duplicated. Why not move this code to `./server/keyshare/db.go`? Then both myirmaserver and tasks can call this.

bobhageman: You where a little too quick, had almost finished it already;) It is now pushed and the whole thingy is ready for review again

ivard (approved):

sanderhollaar (dismissed):

CHANGELOG.md
```diff
@@ -16,7 +20,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Race conditions in database logic of revocation storage
 - `irma scheme verify` not detecting missing files in index 
 - Scheme verification/signing does not reject credentials with invalid revocation settings
-- Write transactions within memory implementation of revocation storage may lead to unintended changes
```
ivard, line 19: This item should not be deleted from the changelog.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
```
ivard, line 9: I think the 'Added' part is still missing.

ivard (approved):

sanderhollaar (approved):

PR #315, opened 2023-06-19
Fix: return error for missing files in index

bobhageman:


sanderhollaar (approved):

schemes.go
```diff
@@ -1204,7 +1204,7 @@ func (scheme *SchemeManager) verifyFiles(conf *Configuration) error {
 			return err
 		}
 		if !exists {
-			continue
+			return errors.Errorf("file %s in index is not found on disk", file)
```
sanderhollaar, line 1207: remove `on disk`

bobhageman: @sanderhollaar , mentioned the 'on disk' part to keep the error messages consistent with the existing error which can happen when the verifying a requestor-scheme here:

https://github.com/privacybydesign/irmago/blob/fix-scheme-verify/schemes.go#L1376:
```
return errors.Errorf("file %s of requestor scheme %s in index but not found on disk", file, scheme.ID), SchemeManagerStatusParsingError
```

What do you suggest: remove both occurrences?

sanderhollaar: hmz I think the `on disk` is meaningless but I understand you want to keep the messages consistent. I'm oke with either solution :-)

bobhageman: Then I'll keep it like this

PR #316, opened 2023-06-19
Fix: return error when credential contains invalid revocation settings

bobhageman:


schemes.go
```diff
@@ -1252,13 +1252,30 @@ func (scheme *SchemeManager) parseCredentialsFolder(conf *Configuration, issuer
...
+			if attr.RevocationAttribute {
+				revAttrs = true
+			}
 		}
+
+		if len(cred.RevocationServers) == 0 && revAttrs {
+			return errors.New("Revocation attribute specified, but RevocationServer(s) is missing")
+		}
+		if len(cred.RevocationServers) > 0 && !revAttrs {
+			return errors.New("RevocationServer(s) specified, but revocation attribute: '<Attribute revocation=\"true\" />' is missing")
```
sanderhollaar, line 1276: remove parenthesis in `RevocationServer(s)`

schemes.go
```diff
@@ -1252,13 +1252,30 @@ func (scheme *SchemeManager) parseCredentialsFolder(conf *Configuration, issuer
...
+				return errors.New(fmt.Sprintf("Attribute %s.%s cannot contain revocation=\"true\". This needs to be a separate attribute: <Attribute revocation=\"true\" />", attr.IssuerID, attr.ID))
+			}
+
+			if attr.RevocationAttribute {
+				revAttrs = true
+			}
 		}
+
+		if len(cred.RevocationServers) == 0 && revAttrs {
+			return errors.New("Revocation attribute specified, but RevocationServer(s) is missing")
```
sanderhollaar, line 1273: remove parenthesis in `RevocationServer(s)`

schemes.go
ivard: Is the issue the other way around also handled somewhere? So that would be `len(cred.RevocationServers) == 0 && revAttrs`.

bobhageman: @ivard: yes it is. Right above it actually, at line [1272](https://github.com/privacybydesign/irmago/pull/316/files#diff-d9a0f717b4668b699fae9ed83cd3c40b204b233359281a673934659068d53fa8R1272)

sanderhollaar (approved):

schemes.go
ivard: Ah of course

ivard (approved):

PR #324, opened 2023-07-27
add mssql (Microsoft SQL Server) as supported db type for revocation (with tests)

ivard:


docker-compose.yml
```diff
@@ -32,6 +32,26 @@ services:
...
+  mysql:
+    image: mysql:8
+    environment:
+      MYSQL_ROOT_PASSWORD: testpassword
+      MYSQL_DATABASE: test
+      MYSQL_USER: testuser
+      MYSQL_PASSWORD: testpassword
+    ports:
+      - 3306:3306
+  mssqlserver:
```
ams-azaaiman, line 44: For consistency I would call this `sqlserver`.

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/go-co-op/gocron v1.28.3
 	github.com/go-errors/errors v1.4.2
 	github.com/go-redis/redis/v8 v8.11.5
-	github.com/golang-jwt/jwt/v4 v4.4.1
+	github.com/golang-jwt/jwt/v4 v4.5.0
```
ams-azaaiman, line 17: Is this required to get sqlserver support in?

docker-compose.yml
```diff
@@ -32,6 +32,26 @@ services:
     # We have to wait until the database is up and running.
     # Database might already be running, so we need to do a cleanup first.
     command: /bin/sh -c "sleep 5 && psql -f cleanup.sql && psql -f schema.sql"
+  mysql:
```
ams-azaaiman, line 35: Would it be an idea to introduce [profiles](https://docs.docker.com/compose/compose-file/15-profiles/)?  
If you spin things up locally, it would be enough to have postgres running.  
For running tests you need mysql and sqlserver too.

go.mod
ivard: That happened automatically when I added the `gorm.io/driver/sqlserver` dependency. That's because they also mention this dependency in their go.sum and we use different versions.

https://github.com/go-gorm/sqlserver/blob/6e2cde390b0a489e530d5771f0885fb138240ba4/go.sum#L10

docker-compose.yml
ivard: There are already some `depends_on` relations in place now, so if you [run the `irma` service](https://github.com/privacybydesign/irmago/blob/2e12f46a8315fc434c17a83d017478ac2e215c91/README.md#running) then `mysql` and `sqlserver` are not started. But indeed, if you'd run `docker-compose up` then all databases would spin up.

In most cases I only use `docker-compose up` if I want to run the tests using the debugger in my IDE, and then you actually want all databases to be up. We can introduce a profile for this and add that [here in the README](https://github.com/privacybydesign/irmago/blob/2e12f46a8315fc434c17a83d017478ac2e215c91/README.md#running-the-tests). I'm not sure whether the performance gain for developers outweighs the inconvenience of having to switch profiles.

ams-azaaiman: Okay, that makes sense. Keep it as it is 👍

bobhageman (approved):

sjoerdlmkns (approved):

PR #325, opened 2023-07-28
feat: custom WrapErrorPrefix that respects SessionError

sjoerdlmkns:


messages.go
```diff
@@ -465,7 +465,7 @@ func ParseRequestorJwt(action string, requestorJwt string) (RequestorJwt, error)
 		return nil, err
 	}
 	if err := retval.RequestorRequest().Validate(); err != nil {
-		return nil, errors.WrapPrefix(err, "Invalid JWT body", 0)
+		return nil, WrapErrorPrefix(err, "Invalid JWT body")
```
ivard, line 468: These occurrences can also occur in irmaclient if I'm right:
image

messages.go
```diff
@@ -513,3 +513,14 @@ type FrontendSessionStatus struct {
 	Status      ServerStatus `json:"status"`
 	NextSession *Qr          `json:"nextSession,omitempty"`
 }
+
+func WrapErrorPrefix(err error, msg string) error {
+	// If error is already a SessionError, just add the prefix to the info
+	if sessionErr, ok := err.(*SessionError); ok {
+		sessionErr.Info = msg + ": " + sessionErr.Info
+		return sessionErr
+	}
```
ivard, line 522: I'd make a new instance of `SessionError`, because the error interface is implemented on `*SessionError`. Now you also modify the struct instance that was passed as `err`.

Furthermore, you can use `fmt.Sprintf` to format the string a bit more nicely.

```
	// If error is already a SessionError, just add the prefix to the info
	if sessionErr, ok := err.(*SessionError); ok {
		return &SessionError{
			Err:         sessionErr.Err,
			ErrorType:   sessionErr.ErrorType,
			Info:        fmt.Sprintf("%s: %s", msg, sessionErr.Info),
			RemoteError: sessionErr.RemoteError,
		}
	}
```

ivard (changes requested):

ivard (dismissed):

ivard (approved):

MartijnYivi (approved):

PR #327, opened 2023-08-09
Feat: use new keyshare protocol in irmaclient

ivard:
In this PR I implemented the clientside parts of the new keyshare protocol. The server endpoints were already introduced in #267, #273, #274 and #278.

This PR does not include introducing a new JWT type for verifiers. This will be done later in a separate PR because it needs some thinking first. There is no API design yet about how we post the JWT to verifiers and we must decide whether we want to keep the difference between the message format in signing sessions and disclosure sessions. The JWT handling for verifiers is not mandatory for using the new keyshare protocol, so this issue is not blocking.

### DIFF
--- a/internal/sessiontest/keyshare_test.go
+++ b/internal/sessiontest/keyshare_test.go
@@ -125,12 +125,19 @@ func TestMultipleKeyshareServers(t *testing.T) {
 	client.KeyshareEnroll(test2SchemeID, nil, "12345", "en")
 	require.NoError(t, <-handler.c)
 
+	// A session request that contains attributes from both test and test2 (both distributed schemes) should fail.
 	request := irma.NewDisclosureRequest(
 		irma.NewAttributeTypeIdentifier("test.test.mijnirma.email"),
 		irma.NewAttributeTypeIdentifier("test2.test.mijnirma.email"),
 	)
-	doSession(t, request, client, irmaServer, nil, nil, nil)
+	_, _, _, err = irmaServer.irma.StartSession(request, nil)
+	require.ErrorIs(t, err, irma.ErrMultipleDistributedSchemes)
 
+	// Do a session with a request that contains attributes from test2 only.
+	request = irma.NewDisclosureRequest(
+		irma.NewAttributeTypeIdentifier("test2.test.mijnirma.email"),
+	)
+	doSession(t, request, client, irmaServer, nil, nil, nil)
 	logs, err = client.LoadNewestLogs(20)
 	require.NoError(t, err)
 	require.Len(t, logs, logsAmount+2)
--- a/internal/sessiontest/legacy_test.go
+++ b/internal/sessiontest/legacy_test.go
@@ -5,6 +5,7 @@ import (
 
 	irma "github.com/privacybydesign/irmago"
 	"github.com/privacybydesign/irmago/internal/test"
+	"github.com/privacybydesign/irmago/internal/testkeyshare"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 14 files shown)


irmaclient/keyshare.go
```diff
@@ -359,34 +361,50 @@ func (ks *keyshareSession) verifyPinAttempt(pin string) (
...
-		pkids[managerID] = append(pkids[managerID], &irma.PublicKeyIdentifier{Issuer: irma.NewIssuerIdentifier(pk.Issuer), Counter: pk.Counter})
+	}
+
+	// TODO: this code is copied from gabi, because there was no way to call the gabi code. Check how we can resolve this.
+	// The secret key may be used across credentials supporting different attribute sizes.
+	// So we should take it, and hence also its commitment, to fit within the smallest size -
+	// otherwise it will be too big so that we cannot perform the range proof showing
+	// that it is not too big.
+	skRandomizer := common.RandomBigInt(new(big.Int).Lsh(big.NewInt(1), gabikeys.DefaultSystemParameters[1024].LmCommit))
+	randomizers := map[string]*big.Int{"secretkey": skRandomizer}
```
ivard, line 388: We should still decide what we want to do with this TODO.

internal/sessiontest/legacy_test.go
```diff
@@ -68,3 +69,13 @@ func TestWithoutPairingSupport(t *testing.T) {
...
+
+func TestLinkableKeyshareResponse(t *testing.T) {
+	keyshareServer := testkeyshare.StartKeyshareServer(t, logger, irma.NewSchemeManagerIdentifier("test"))
+	defer keyshareServer.Stop()
+	client, handler := parseStorage(t, optionLinkableKeyshareResponse)
+	defer test.ClearTestStorage(t, client, handler.storage)
+	irmaServer := StartIrmaServer(t, nil)
+	defer irmaServer.Stop()
+	keyshareSessions(t, client, irmaServer)
+}
```
ivard, line 81: This test does not work yet. This means that the new irmaclient does not support older IRMA server. This is an issue that needs to be resolved. I validated the bug by doing a session using an actual older IRMA server, so I think we can rule out a component test issue.

My guess would be that something goes wrong when the prooflist is being converted to make them work with the linkable keyshare JWTs. This happens in the function `removeKeysharePsFromProofUs`.

internal/sessiontest/keyshare_test.go
```diff
@@ -125,12 +125,19 @@ func TestMultipleKeyshareServers(t *testing.T) {
...
 	require.NoError(t, <-handler.c)
 
+	// A session request that contains attributes from both test and test2 (both distributed schemes) should fail.
 	request := irma.NewDisclosureRequest(
 		irma.NewAttributeTypeIdentifier("test.test.mijnirma.email"),
 		irma.NewAttributeTypeIdentifier("test2.test.mijnirma.email"),
 	)
-	doSession(t, request, client, irmaServer, nil, nil, nil)
+	_, _, _, err = irmaServer.irma.StartSession(request, nil)
+	require.ErrorIs(t, err, irma.ErrMultipleDistributedSchemes)
```
ivard, line 134: Gabi seems to have no support for doing sessions using multiple keyshare servers. This is not blocking for now, but if someone knows a solution that would be nice.

internal/sessiontest/legacy_test.go
ivard: Fixed in #333

internal/sessiontest/keyshare_test.go
ivard: We decided to not implement this for now.

w-ensink (approved):

PR #344, opened 2023-09-07
Improve: send expiry mail when user has only valid addresses or to an…

bobhageman:
…y valid when revalidation disabled

### DIFF
--- a/server/keyshare/tasks/tasks.go
+++ b/server/keyshare/tasks/tasks.go
@@ -128,21 +128,17 @@ func (t *taskHandler) sendExpiryEmails(ctx context.Context, id int64, username, 
 			}
 
 			if err := keyshare.VerifyMXRecord(email); err != nil {
-				if !t.revalidateMail {
-					// It could be this user has another, valid email address. Therefore we do not abort
-					// since is a chance we can still inform the user.
-					return nil
+				if t.revalidateMail {
+					if revErr := t.db.ExecUserContext(ctx, "UPDATE irma.emails SET revalidate_on = $1 WHERE id = $2",
+						time.Now().AddDate(0, 0, 5).Unix(),
+						id); revErr != nil {
+						t.conf.Logger.WithField("error", revErr).Error("Could not update email address to set revalidate_on")
+						return revErr
+					}
 				}
 
-				if revErr := t.db.ExecUserContext(ctx, "UPDATE irma.emails SET revalidate_on = $1 WHERE id = $2",
-					time.Now().AddDate(0, 0, 5).Unix(),
-					id); revErr != nil {
-					t.conf.Logger.WithField("error", revErr).Error("Could not update email address to set revalidate_on")
-					return revErr
-				}
-
-				// Abort to prevent sending mail until the invalid email address is revalidated
-				// so we can send the expiry mail to all, and only valid, addresses at once
+				// Abort to prevent sending mail. If revalidation is enabled, after the e-mail address is revalidated,
+				// the process will continue so an expiry mail is sent to all, and only valid, addresses at once.
 				return err
 			}
 
--- a/server/keyshare/tasks/tasks.go
+++ b/server/keyshare/tasks/tasks.go
@@ -128,16 +128,20 @@ func (t *taskHandler) sendExpiryEmails(ctx context.Context, id int64, username, 
 			}
 
 			if err := keyshare.VerifyMXRecord(email); err != nil {
-				if t.revalidateMail {
-					if err = t.db.ExecUserContext(ctx, "UPDATE irma.emails SET revalidate_on = $1 WHERE id = $2",
-						time.Now().AddDate(0, 0, 5).Unix(),
-						id); err != nil {
-						t.conf.Logger.WithField("error", err).Error("Could not update email address to set revalidate_on")
-						return err
-					}
+				if !t.revalidateMail {
+					// It could be this user has another, valid email address. Therefore we do not abort
+					// since is a chance we can still inform the user.
+					return nil
 				}
 
-				// We wait with further processing until the email address is revalidated
+				if revErr := t.db.ExecUserContext(ctx, "UPDATE irma.emails SET revalidate_on = $1 WHERE id = $2",
+					time.Now().AddDate(0, 0, 5).Unix(),
+					id); revErr != nil {
+					t.conf.Logger.WithField("error", revErr).Error("Could not update email address to set revalidate_on")
+					return revErr
+				}
+
+				// Abort to prevent sending mail until the invalid email address is revalidated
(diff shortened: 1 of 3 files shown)


server/keyshare/tasks/tasks.go
```diff
@@ -128,16 +128,20 @@ func (t *taskHandler) sendExpiryEmails(ctx context.Context, id int64, username,
...
-					if err = t.db.ExecUserContext(ctx, "UPDATE irma.emails SET revalidate_on = $1 WHERE id = $2",
-						time.Now().AddDate(0, 0, 5).Unix(),
-						id); err != nil {
-						t.conf.Logger.WithField("error", err).Error("Could not update email address to set revalidate_on")
-						return err
-					}
+				if !t.revalidateMail {
+					// It could be this user has another, valid email address. Therefore we do not abort
+					// since is a chance we can still inform the user.
+					return nil
```
ivard, line 134: Why was this change necessary? If you return the error, then [this check](https://github.com/privacybydesign/irmago/pull/344/files#diff-d2930c933338a39b361fd236af2fe82f5741b7a689ef97a330534c1bb8cd2b58R229) ensures that the user is skipped, as we want.

server/keyshare/tasks/tasks.go
```diff
@@ -128,16 +128,20 @@ func (t *taskHandler) sendExpiryEmails(ctx context.Context, id int64, username,
...
+				if revErr := t.db.ExecUserContext(ctx, "UPDATE irma.emails SET revalidate_on = $1 WHERE id = $2",
+					time.Now().AddDate(0, 0, 5).Unix(),
+					id); revErr != nil {
+					t.conf.Logger.WithField("error", revErr).Error("Could not update email address to set revalidate_on")
+					return revErr
+				}
+
+				// Abort to prevent sending mail until the invalid email address is revalidated
 				// so we can send the expiry mail to all, and only valid, addresses at once
 				return err
```
ivard, line 146: The problem here was that `err = t.db.ExecUserContext(ctx, "UPDATE irma.emails SET revalidate_on = $1 WHERE id = $2"` overrides the existing error with `nil`. This means that this function returned nil instead of an error if the revalidation was executed.

server/keyshare/tasks/tasks.go
```diff
@@ -223,13 +226,13 @@ func (t *taskHandler) expireAccounts(ctx context.Context) {
...
 					if !t.revalidateMail {
-						// To have the exact same behavior as before email revalidation functionality,
-						// we return nil when the error is ErrInvalidEmail. Additionally we log the error
-						t.conf.Logger.WithField("error", err).Errorf("User decreases processable amount in expireAccounts, id: %d", id)
+						// We can't do anything about an invalid email address or domain
+						// when revalidation is disabled so we log the error and continue iterating
+						t.conf.Logger.WithField("error", err).Errorf("User decreases processable amount in expireAccounts, user_id: %d", id)
 					}
-					return nil
+					return nil // Don't abort the entire task over an email seding issue, only skip this user
```
ivard, line 235: I tested this code locally by adapting one of the unit tests. Maybe it's nice to include this test.

```
func TestEmailRevalidation(t *testing.T) {
	SetupDatabase(t)
	testdataPath := test.FindTestdataFolder(t)
	defer TeardownDatabase(t)

	db, err := sql.Open("pgx", test.PostgresTestUrl)
	require.NoError(t, err)

	th, err := newHandler(&Configuration{
		DBConnStr:   test.PostgresTestUrl,
		DeleteDelay: 30,
		ExpiryDelay: 365,
		EmailConfiguration: keyshare.EmailConfiguration{
			EmailServer:     "localhost:1025",
			EmailFrom:       "test@example.com",
			DefaultLanguage: "en",
		},
		DeleteExpiredAccountFiles: map[string]string{
			"en": filepath.Join(testdataPath, "emailtemplate.html"),
		},
		DeleteExpiredAccountSubjects: map[string]string{
			"en": "testsubject",
		},
		Logger: irma.Logger,
	})
	require.NoError(t, err)

	createUser(t, db, 1, "TemporaryInvalid", time.Now().AddDate(-1, -1, -1).Unix(), []string{"user_1@temporaryinvalidaddress.com"})
	createUser(t, db, 2, "PermanentInvalid", time.Now().AddDate(-1, -1, -1).Unix(), []string{"user_2@permenantlyinvalidaddress.com"})
	createUser(t, db, 3, "PartlyInvalid", time.Now().AddDate(-1, -1, -1).Unix(), []string{"user_3@permenantlyinvalidaddress.com", "user_3@github.com"})

	require.NoError(t, runWithTimeout(th.expireAccounts))
	require.NoError(t, runWithTimeout(th.revalidateMails))

	// Is revalidate_on set for all email addresses?
	assert.Equal(t, 3, countRows(t, db, "emails", "revalidate_on IS NOT NULL"))
	assert.Equal(t, 0, countRows(t, db, "users", "delete_on IS NOT NULL"))

	// Correct the temporary invalid e-mail address to a valid one, simulating it becoming valid, and 'forward time'
	_, err = db.Exec("UPDATE irma.emails SET email = $1, revalidate_on = $2 WHERE id = $3", "user_1@github.com", time.Now().AddDate(0, 0, -1).Unix(), 1)
	require.NoError(t, err)

	// Forward time for the invalid address
	_, err = db.Exec("UPDATE irma.emails SET revalidate_on = $1 WHERE email LIKE $2", time.Now().AddDate(0, 0, -1).Unix(), "%@permenantlyinvalidaddress.com")
	require.NoError(t, err)

	// Next revalidation run: the invalid email address should be deleted now
	require.NoError(t, runWithTimeout(th.revalidateMails))

	assert.Equal(t, 2, countRows(t, db, "emails", ""))

	// Next expireAccounts run: the PartlyInvalid user should be expired since the permanently invalid address is deleted now.
	require.NoError(t, runWithTimeout(th.expireAccounts))
	assert.Equal(t, 3, countRows(t, db, "users", ""))
	assert.Equal(t, 2, countRows(t, db, "users", "delete_on IS NOT NULL"))
}
```

server/keyshare/tasks/tasks.go
bobhageman: By using `revErr,` `err` is not overwritten with `nil` so it does return `keyshare.ErrInvalidEmail` and/or `keyshare.ErrInvalidEmailDomain` and any other type of exception to `expireAccounts(..)`. I agree this it not the most beautiful pattern, but did not see another way to do it without a much bigger refactoring.

server/keyshare/tasks/tasks.go
bobhageman: Nice addition!

ivard (dismissed):

ivard (approved):

sjoerdlmkns (approved):

PR #355, opened 2023-10-24
Add irma cmd option to prolong keys

bobhageman:


ivard:
Not needed anymore. Hopefully all issuers prolong their key in time and otherwise it can be done manually.

PR #358, opened 2023-11-21
Check MX record hostname validity

bobhageman:


server/keyshare/email.go
```diff
@@ -150,21 +150,33 @@ func VerifyMXRecord(email string) error {
...
+	for _, h := range records {
+		// Check if host specified at MX record valid
+		if addr, err := net.LookupHost(h.Host); err != nil || len(addr) == 0 {
+			invalidHosts++
+		}
+	}
+
+	if invalidHosts >= len(records) {
+		return ErrInvalidEmailDomain
+	}
```
ivard, line 179: You assumed here that the mail relay does not return an error if at least one of the hosts in the MX has a valid value? Do we know this for sure? Otherwise, it could be much simpler and we could already do an early return if at least one of the hosts is invalid.

server/keyshare/email.go
```diff
@@ -150,21 +150,34 @@ func VerifyMXRecord(email string) error {
...
 		if records, err := net.LookupIP(host); err != nil || len(records) == 0 {
 			return ErrInvalidEmailDomain
 		}
-		return nil
 	}
+
+	hasValidHost := false
+	for _, h := range records {
+		// Check if host specified at MX record is valid
+		if addr, err := net.LookupHost(h.Host); err == nil || len(addr) > 0 {
```
ivard, line 172: This has to be `&&` if I'm right.

server/keyshare/myirmaserver/server.go
```diff
@@ -311,10 +311,10 @@ func (s *Server) handleEmailLogin(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// In case sendLoginEmail fails with errEmailNotFound or errTooManyRequests, then we
+	// In case sendLoginEmail fails with errEmailNotFound, errTooManyRequests or ErrInvalidEmailDomain, then we
 	// should not write an error. Otherwise, we would leak information about our user base.
 	err := s.sendLoginEmail(r.Context(), request)
-	if err != nil && err != errEmailNotFound && err != errTooManyTokens {
+	if err != nil && err != errEmailNotFound && err != errTooManyTokens && err != keyshare.ErrInvalidEmailDomain {
```
ivard, line 317: What's exactly the issue about returning a 400 when the domain is not valid?

ivard (changes requested):

server/keyshare/email.go
bobhageman: Good one! You are absolutely right.

server/keyshare/myirmaserver/server.go
bobhageman: Double checked and you're right indeed! It does not cause an issue (in the way of leaking information) if we would return the 400 at a properly formatted but non-working email address. I'll reverse that bit and correct the test accordingly.

ivard (dismissed):

ivard (approved):

sanderhollaar (approved):

PR #359, opened 2023-11-27
Fix: cookies not stored in irmaclient when received from a Set-Cookie…

ivard:
… header


ivard:
Test failed due to #360

bobhageman (approved):

sanderhollaar (approved):

PR #391, opened 2024-11-08
Implement signature verification api

DeD1rk:
Based on a discussion from a long time ago (https://yiviapp.slack.com/archives/C0KCTQ0BC/p1713794023128249) I decided to implement an API endpoint for verifying signatures in the irma server.

Currently, there's no authorization: as the user already has a signature, and authenticator does not prevent doing disclosures with excessive personal data. 

However, maybe it should still be implement to prevent freeloading on a public irma server. I'm not sure about that, nor about how that would look. Alternatively, it could also be made configurable whether the endpoint should be available at all. That way, it can be disabled by default, and hosting parties who are afraid of people using it without authentication can still turn it on when reachable only from a private network.

I haven't made docs or implemented implemented any tests (yet). The verification logic is already tested though, and I'm using this without any issues so far in my project. 

Let me know if you have any ideas or requests! If you guys are interested in merging this, I can probably take a look at unittests and docs some time.

### DIFF
--- a/requests.go
+++ b/requests.go
@@ -19,14 +19,15 @@ import (
 )
 
 const (
-	LDContextDisclosureRequest      = "https://irma.app/ld/request/disclosure/v2"
-	LDContextSignatureRequest       = "https://irma.app/ld/request/signature/v2"
-	LDContextIssuanceRequest        = "https://irma.app/ld/request/issuance/v2"
-	LDContextRevocationRequest      = "https://irma.app/ld/request/revocation/v1"
-	LDContextFrontendOptionsRequest = "https://irma.app/ld/request/frontendoptions/v1"
-	LDContextClientSessionRequest   = "https://irma.app/ld/request/client/v1"
-	LDContextSessionOptions         = "https://irma.app/ld/options/v1"
-	DefaultJwtValidity              = 120
+	LDContextDisclosureRequest            = "https://irma.app/ld/request/disclosure/v2"
+	LDContextSignatureRequest             = "https://irma.app/ld/request/signature/v2"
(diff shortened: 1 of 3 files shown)


saravahdatipour:
The attribute-based signature verification endpoint seems like a nice idea.
The problem with having the endpoint without authentication is that, any public irma server then will be able to process personal data through the verification of signatures, and the irma server has no control over whose data it is processing, which becomes a GDPR problem.
It is also possible to use the existing session authentication mechanisms to authenticate attribute-based signature verification sessions as well. 
This is mentioned in the docs [here](https://irma.app/docs/irma-server/#requestor-authentication ) 
In irmago, session authentication can be done using any of the structs that implements Authenticator interface.
You can see it in use for example, in https://github.com/privacybydesign/irmago/blob/52fd9f2093752bfe44cfa453e74a8c6f29ff4924/server/requestorserver/server.go#L247-L277

DeD1rk:
@saravahdatipour so you'd like me to take a look at implementing some form of authorization?

Would perhaps a simple boolean 'enable signature verification' config option work? There's no way for irmago to know what kind of attributes are in the signature (or what personal information is in the message) before receiving (and hence processing) it, so fine-grained permissions are not possible here. 
For my specific use-case, it is perfectly acceptable to process arbitrary personal data from GDPR perspective, so enabling it (despite using a publicly reachable irma server) would be just fine for me, and I imagine that other uses of it (such as a current thesis project for PubHubs) would also need public access to it, or can just use it on a server that is not publicly reachable. It could be disabled by default of course.

I can see some added value in having a complete requestor authentication system, but as it doesn't quite fit in the existing requestor authentication for session requests, I think that would be a pretty big addition, that's a little beyond my skillset and the time I have.

PR #392, opened 2024-11-08
Add arm64 docker build to delivery workflow

DeD1rk:
This switches to the build-push-action with buildx to build for multiple architectures in parallel.

This is nice if you want to run an irma server on a Raspberry Pi (if you're a student self-hosting it at home) or e.g. on AWS ec2 t4g instances.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Internal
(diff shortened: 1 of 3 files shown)


ivard:
Is the docker build-push-action smart enough that it doesn't build the Docker container twice when building for `edge` or `latest`?

DeD1rk:
It is. To demo, I triggered it on my fork again, where I think you should be able to check the logs now: https://github.com/DeD1rk/irmago/actions/runs/13791508731/job/38572364363. However, I've not added any further caching (of intermediate layers). That's pretty easy when using the docker action, and can help save a lot of time, but really isn't the purpose of this PR.

image

ivard (approved):
Ah nice, the failing test is a non-related issue that's fixed on `master`. I'll update the branch when we merge to prevent superfluous merge commits.

Issue #394, opened 2024-12-13
Redis WAIT commands being applied in MULTI transaction without checking number of replicas


ivard:
In the Redis documentation the following remark is stated:

> If the command is sent as part of a [MULTI](https://redis.io/docs/latest/commands/multi/) transaction (since Redis 7.0, any context that does not allow blocking, such as inside scripts), the command does not block but instead just return ASAP the number of replicas that acknowledged the previous write commands.

We currently do this in an attempt to implement a best-effort approach for consistency in Sentinel setups. This can be found in the [Redis documentation](https://redis.io/docs/latest/operate/rs/databases/durability-ha/consistency/). If the command does not block in MULTI until the write is acknowledged by two replicas, then the WAIT has little effect. Therefore, we should consider checking the number of replicas the WAIT command returns and implement an alternative flow if the number is less than 2.

If this change is considered to have a too small effect, then the WAIT commands might be superfluous at all.

Examples of places where the WAIT command is called:
https://github.com/privacybydesign/irmago/blob/52fd9f2093752bfe44cfa453e74a8c6f29ff4924/server/irmaserver/sessions.go#L275
https://github.com/privacybydesign/irmago/blob/52fd9f2093752bfe44cfa453e74a8c6f29ff4924/server/irmaserver/sessions.go#L352
https://github.com/privacybydesign/irmago/blob/52fd9f2093752bfe44cfa453e74a8c6f29ff4924/server/keyshare/myirmaserver/session.go#L107
https://github.com/privacybydesign/irmago/blob/52fd9f2093752bfe44cfa453e74a8c6f29ff4924/server/keyshare/myirmaserver/session.go#L153

DibranMulder:
What should we do other then raising an error? I suggest we log the amount of acknowledged replicas and inspect the behavior. If the systems behaves as expected we might raise an error in the future.

ivard:
> What should we do other then raising an error? I suggest we log the amount of acknowledged replicas and inspect the behavior. If the systems behaves as expected we might raise an error in the future. 

What the WAIT should do is to block until the data is written to at least 2 replicas. We run 3 so even if one's down, it should be possible to reach 2. If only one replica acknowledges, it means the application is communicating with an orphan replica which mostly means something is wrong and the data in Redis cannot be trusted.

This issue was about the documention stating that WAIT's inside a MULTI may not work as expected. The wait should probably just be moved after the EXEC that finishes the MULTI instead of before if done so in the code.

PR #397, opened 2025-01-30
Log the amount of replica's Redis wrote to

DibranMulder:
This pull request includes several changes to improve logging in the `redisSessionStore` methods across different files. The main focus is on enhancing the visibility of Redis datastore acknowledgments by logging the number of replicas that acknowledged the write.

Deploying this change should contribute to fixing #394


ivard:
In #398 I have a different proposal to just fully remove the waits.

PR #401, opened 2025-03-06
Return from sse loop when sseclient.Notify() fails

w-ensink:
To prevent infinite loop that consumes all of the CPU.

Related to https://github.com/privacybydesign/irmamobile/issues/325

### DIFF
--- a/internal/sessiontest/revocation_test.go
+++ b/internal/sessiontest/revocation_test.go
@@ -1,6 +1,9 @@
 //go:build !local_tests
 // +build !local_tests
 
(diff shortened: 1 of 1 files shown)


DibranMulder (dismissed):
Really nice find

ivard (dismissed):
Maybe it's nice to add an entry to the unreleased part of the CHANGELOG.md. Otherwise, we might forget mentioning it when we are releasing.

ivard (approved):

saravahdatipour (approved):

PR #422, opened 2025-05-06
Fix for CVE GHSA-pv8v-c99h-c5q4

kamphuisem:
This pull request introduces significant updates to the session testing framework, primarily to support requestor server functionality and improve flexibility in handling chained sessions and server configurations. It includes changes to test helper functions, session handling logic, and the addition of new test cases for unauthorized and non-requestor chained sessions.

### Enhancements to Session Handling:
* Updated `startServer` to accept an additional `requestorServer` parameter, allowing tests to use either an `IrmaServer` or a `requestorserver.Server`. This ensures compatibility with requestor server features.
* Modified `startSessionAtServer` to handle sessions for both server types (`IrmaServer` and `requestorserver.Server`), improving test coverage for requestor server scenarios.
* Enhanced `waitSessionFinished` to support session result retrieval from both server types, ensuring consistent behavior across different configurations.

### New Test Cases:
* Added `doUnauthorizedChainedSession` to test scenarios where chained sessions are not authorized by the requestor server configuration.
* Introduced `doNonRequestorChainedSessions` to validate behavior for chained sessions unsupported by requestor servers, ensuring proper error handling.

### Refactoring and Code Simplification:
* Refactored `doSession` and `doChainedSessions` to include the `requestorServer` parameter, aligning the function signatures with the updated server logic. [[1]](diffhunk://#diff-9873bc8ff31f54f611449e25b3273fe729a82d793970e88226343fd4b4eb2734R278) [[2]](diffhunk://#diff-9873bc8ff31f54f611449e25b3273fe729a82d793970e88226343fd4b4eb2734R349-R355)
* Updated `chainedServerHandler` and `StartNextRequestServer` to use explicit parameters (`publicKey`, `credentialTypes`) instead of relying on the server configuration object, improving modularity. [[1]](diffhunk://#diff-68ffaae070a6bbe0623df8506dbf61f5e735898da6762f2c49f987e86f9b35cdL167-R171) [[2]](diffhunk://#diff-68ffaae070a6bbe0623df8506dbf61f5e735898da6762f2c49f987e86f9b35cdL204-R209)

### Configuration Enhancements:
* Added `RequestorServerPermissionsConfiguration` to define requestor-specific permissions for disclosure and issuance, enabling more granular control over requestor server capabilities.

### Minor Changes:
* Adjusted test cases in `keyshare_test.go` to include the new `requestorServer` parameter where applicable, ensuring compatibility with the updated `doSession` function. [[1]](diffhunk://#diff-96d78f72014382fb8ba8bf5cfe9f16a9712a3a51f89f824c142c68b37b2c7086L55-R55) [[2]](diffhunk://#diff-96d78f72014382fb8ba8bf5cfe9f16a9712a3a51f89f824c142c68b37b2c7086L83-R87) [[3]](diffhunk://#diff-96d78f72014382fb8ba8bf5cfe9f16a9712a3a51f89f824c142c68b37b2c7086L123-R131)


server/irmaserver/api.go
```diff
@@ -27,13 +27,14 @@ import (
...
-	activeSSEHandlers      map[irma.RequestorToken]bool
-	activeSSEHandlersMutex sync.Mutex
+	conf                     *server.Configuration
+	router                   *chi.Mux
+	sessions                 sessionStore
+	scheduler                *gocron.Scheduler
+	serverSentEvents         *sse.Server
+	activeSSEHandlers        map[irma.RequestorToken]bool
+	activeSSEHandlersMutex   sync.Mutex
+	NextSessionAuthorization irma.SessionAuthorizationFunc
```
DibranMulder, line 37: i don't think the nextSessionAuthorization should be a property of the server.

server/requestorserver/server.go
```diff
@@ -141,6 +150,10 @@ func New(config *Configuration) (*Server, error) {
 	if err := config.initialize(); err != nil {
 		return nil, err
 	}
+
+	// Authorize NextSessions against the requestor configuration
```
DibranMulder, line 154: I think this is prone to raise conditions, since it overwrites server state while it should be bound to session/request scope.

DibranMulder (changes requested):

server/requestorserver/server.go
kamphuisem: It is a 'function pointer' which is overwritten on the server startup.
This will be done once, so it will use a different function for all sessions using the requestorserver (instead of plain IRMA server and/or keyshare).
There is no race condition possible with that.

w-ensink: I think the only possible race condition would be changing the permissions in the config while checking permissions. That's not something that will happen, as the permissions are read from a config file and never changed. I don't think it will use a different function for each session tho. It's a pointer to the same function of the config passed in at creation every time.

w-ensink:
What is the reason for not putting all required fields for checking permissions into the `server.Configuration` struct?
This struct is flattened and embedded in `requestorserver.Configuration`, so it would still be available to the requestor server too that way, right?

.vscode/launch.json
```diff
@@ -8,7 +8,7 @@
             "mode": "auto",
             "program": "${workspaceFolder}/irma/main.go",
             "args": [
-                "server"
+                "server",
```
confiks, line 11: Why this change? (it's invalid JSON but maybe vscode support JSON5)

confiks (approved):

w-ensink (dismissed):

w-ensink (approved):

confiks (approved):

w-ensink:
> Still think the next NextSessionAuthorization should not be a property of the server struct though.

Yeah I think we can all agree on that. It takes quite some more work to really do it right tho. Hopefully we can fix it the right way somewhere in the future.

PR #439, opened 2025-05-14
Update transport.go to allow usage of http proxies

### LINKED ISSUE #423: Use http-proxy for outgoing connections
When we try to start irmago/yivi in our lab/testing environment, we run into the issue that on startup, it will try to download various links (example: https://privacybydesign.foundation/schememanager/irma-demo/description.xml). However, our networks do not have unlimited access to the public internet, as every connection needs to go thru a proxy-server.

Description: GET https://privacybydesign.foundation/schememanager/irma-demo/description.xml giving up after 2 attempt(s): con>

We've looked and tried setting default HTTP_PROXY/HTTPS_PROXY env vars, but these don't seem to be used.

Feature request: add support for http/https proxies so the application can be started in 'secured' environments.

sigio:
By setting `http.Transport.Proxy = http.ProxyFromEnvironment`, irmago will respect system proxy settings.
This allows use in secured or firewalled environments requiring a proxy server for outgoing HTTP(S) traffic.

This code compiles, and seems to use a proxy when setting HTTPS_PROXY

```
May 14 21:49:29 t480 tinyproxy[331565]: Request (file descriptor 12): CONNECT schemes.yivi.app:443 HTTP/1.1
May 14 21:49:29 t480 tinyproxy[331565]: No upstream proxy for schemes.yivi.app
May 14 21:49:29 t480 tinyproxy[331565]: opensock: opening connection to schemes.yivi.app:443
```

```
./irma.bin scheme download
No irma_configuration path specified, using /home/sigio/.local/share/irma/irma_configuration
[21:48:53.597714]  INFO downloading default schemes (may take a while)
[21:48:53.829394]  INFO checking for updates scheme=irma-demo type=issuer
[21:48:54.055701]  INFO scheme is outdated, updating scheme=irma-demo type=issuer
[21:49:32.966370]  INFO checking for updates scheme=pbdf type=issuer
[21:49:33.208570]  INFO scheme is outdated, updating scheme=pbdf type=issuer
[21:49:49.904658]  INFO Finished downloading schemes

```
Fixes #423

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Fixed
+- `irma.HTTPTransport` now honours the `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` environment variables, so irmago can run where outgoing traffic must pass through a proxy; it was the only outbound transport in irmago that ignored them. Operators with a system-wide proxy can keep internal traffic (requestor callbacks, keyshare and revocation servers) off it with `NO_PROXY`; loopback addresses are never proxied ([#423](https://github.com/privacybydesign/irmago/issues/423))
 
 ## [1.3.0] - 2026-08-12
 ### Added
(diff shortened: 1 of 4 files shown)


kamphuisem:
We need to verify that this change will also work with other parts of the IRMA server, because the transport is used for several parts of the IRMA server.
For example, I don't know if proxies will handle SSE correctly with default configurations.
This could also be a breaking change for existing installations, where proxies are configured on a system/user level, but are now not being used by our server.

sigio:
Patch against current master branch

rubenhensen (approved):

kamphuisem (approved):

PR #457, opened 2025-06-10
Add batch issuance of SD-JWTs

kamphuisem:


server/irmaserver/helpers.go
```diff
@@ -826,37 +830,65 @@ func (session *sessionData) generateSdJwts(privKey *ecdsa.PrivateKey, issuerUrl
...
+	var index uint = 0
 	for _, cred := range issuanceReq.Credentials {
-		disclosures, err := sdjwtvc.MultipleNewDisclosureContents(cred.Attributes)
-		if err != nil {
-			return nil, err
+		credentialType := cred.CredentialTypeID.String()
+
+		// We want each credential to have the same issuance and expiration dates, so we use a static clock
+		systemClock := sdjwtvc.NewSystemClock()
+		staticClock := sdjwtvc.StaticClock{CurrentTime: systemClock.Now()}
```
w-ensink, line 842: There are 2 different clocks here, is that intentional?

w-ensink (approved):

server/irmaserver/helpers.go
kamphuisem: Yes; for each credential I'm using the same 'timestamp', so each Credential Info object will be the same for each credential.
This can probably be removed when we change from IRMA to OpenID4VCI protocol, because I think we will not use the CredentialInfo object at that point.

Issue #455, opened 2025-06-06
Add Key-Binding pub key to issuance


kamphuisem:
A complete SD-JWT VC with Keybinding JWT requires the issuer to add a pub key for the holder in the `cnf` field of the issuer signed JWT. Since we're adding SD-JWT VC issuance to IRMA, the irma server also needs access to this public key. We can probably send in somewhere around here:
https://github.com/privacybydesign/irmago/blob/3c1417dd765f159fd5b4f17423971844b8a05131/irmaclient/session.go#L562

What we do need to keep in mind is that we probably want some kind of proof of association between the public key and the source of the public key (probably the keyshare server). This can be done by adding a signature to the holder public key and verifying that agains the public key of the keyshare server. The question then becomes whether we want that for each pub key or for each batch of pub keys (e.g. packed per 10 or 50).

The requestor server sends a request to the irma server with an `IssuanceRequest` that contains `CredentialRequest` objects.
The `CredentialRequest` objects contain optional info about the number of sdjwt instances to be created.

The app obtains this `IssuanceRequest` before it makes a request to the `/irma/session/commitments` endpoint. At this point the app should therefore
create/gather enough key binding pub keys for all the sdjwts to be created by the irma server.

The app then makes a request containing an `IssueCommitmentMessage`, in which it should put the list of pub keys to be used by the irma server.

If the session doesn't use a keyshare server (e.g. a demo attribute) then it will use `session.getProof()` -> `client.IssueCommitments()`.

If the session does use a keyshare server it will be created in `session.KeyshareDone()`.

PR #462, opened 2025-06-13
Add key binding jwt support to SD-JWT VC issuance over IRMA

w-ensink:
Relates to #455 and https://github.com/privacybydesign/irmamobile/issues/357

### DIFF
--- a/irmaclient/dcql_test.go
+++ b/irmaclient/dcql_test.go
@@ -680,7 +680,7 @@ func testDcqlSatisfiableSingleCredentialSingleOption(t *testing.T) {
 // ===========================================================================================
 
 func createSdJwtAndInfo(t *testing.T, keyBinder sdjwtvc.KeyBinder, credentialId string, attributes map[string]string) (sdjwtvc.SdJwtVc, *irma.CredentialInfo) {
-	sdjwt, err := createSdJwtVc(keyBinder, credentialId, "https://openid4vc.staging.yivi.app", attributes)
+	sdjwt, err := createTestSdJwtVc(keyBinder, credentialId, "https://openid4vc.staging.yivi.app", attributes)
 	require.NoError(t, err)
 
 	info, _, err := createCredentialInfoAndVerifiedSdJwtVc(sdjwt, sdjwtvc.CreateDefaultVerificationContext())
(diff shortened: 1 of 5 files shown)


messages.go
```diff
@@ -302,6 +302,9 @@ type DisclosedAttributeIndex struct {
 type IssueCommitmentMessage struct {
 	*gabi.IssueCommitmentMessage
 	Indices DisclosedAttributeIndices `json:"indices,omitempty"`
+	// Should be an instance of jwk.Key,
+	// but that was not possible because that can't be deserialized automatically.
+	KeyBindingPubKeys []json.RawMessage `json:"keybinding_pub_keys,omitempty"`
```
kamphuisem, line 307: Isn't the 'usual' way to use `MarshalJSON` and `UnmarshalJSON` functions for this, instead of using `json.RawMessage` and `irmaserver/handle.go/parseKeyBindingPubKeys` function?

server/irmaserver/helpers.go
```diff
@@ -811,7 +808,7 @@ func (s *Server) newSession(
 	return ses, nil
 }
 
-func (session *sessionData) generateSdJwts(privKey *ecdsa.PrivateKey, issuerUrl string, allowNonHttps bool) ([]sdjwtvc.SdJwtVc, error) {
+func (session *sessionData) generateSdJwts(privKey *ecdsa.PrivateKey, kbPubKeys []jwk.Key, issuerUrl string, allowNonHttps bool) ([]sdjwtvc.SdJwtVc, error) {
```
kamphuisem, line 811: Maybe good to add a check (before start generating) to validate `irma.CalculateAmountOfSdJwtsToIssue(issuanceReq)` matches `len(kbPubKeys)` ?

eudi/credentials/sdjwtvc/kbjwt.go
```diff
@@ -31,19 +35,61 @@ const (
...
-	CreateKbJwt(hash string, holderKey jwk.Key, nonce string, audience string) (KeyBindingJwt, error)
+	CreateKeyBindingJwt(hash string, holderPubKey jwk.Key, nonce string, audience string) (KeyBindingJwt, error)
 }
 
-type DefaultKbJwtCreator struct {
-	Clock      Clock
-	JwtCreator JwtCreator
+type DefaultKeyBinder struct {
+	Clock Clock
+	Keys  map[string]*ecdsa.PrivateKey
```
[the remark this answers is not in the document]
w-ensink: Yeah good point. These values shouldn't be used outside of this struct anyway.

messages.go
w-ensink: We could manually implement an unmarshal function for `IssueCommitmentMessage`, but the regular Unmarshal doesn't work with `jwk.Key` since that's an interface.

server/irmaserver/helpers.go
w-ensink: Good one

PR #463, opened 2025-06-16
Add issuer certificate chain for SD-JWT VC issuance to irma server

w-ensink:


irma/cmd/config.go
```diff
@@ -42,6 +42,25 @@ func configureEmail() keyshare.EmailConfiguration {
 	}
 }
 
+func getSdJwtIssuanceConfigFromCli() *server.SdJwtIssuanceSettings {
```
kamphuisem, line 45: I believe these settings don't have to come from CLI perse (as viper can read from all kinds of sources).
Maybe change to `readSdJwtIssuanceConfig`

kamphuisem (approved):

irma/cmd/config.go
w-ensink: That would also imply reading from a file, while this is (I think) purely using either the CLI flags or the environment variables

PR #473, opened 2025-07-31
Add credential count for SdJwts

w-ensink:
Relates to https://github.com/privacybydesign/irmamobile/issues/357


irmaclient/sdjwtvc_storage.go
```diff
@@ -15,6 +15,23 @@ import (
 
 // ========================================================================
 
+type SdJwtMetadata struct {
```
w-ensink, line 18: Combine the two and use the instance count as a reference for how many there were initially

requests.go
```diff
@@ -806,6 +808,13 @@ func (ir *IssuanceRequest) GetCredentialInfoList(
 			if err != nil {
 				return nil, err
 			}
+
+			// in this case we'll assume the server is capable to build this
+			// so we just put in the usual default amount
+			if info.InstanceCount == nil && ir.RequestSdJwts {
+				defaultAmount := DefaultSdJwtIssueAmount
+				info.InstanceCount = &defaultAmount
```
w-ensink, line 816: Fix the SessionRequest to have only a credential count for each requested credential, no more boolean.

w-ensink: Should probably happen in a new branch, because it will break a lot of things. A new sms-issuer will need to be built for example.

irmaclient/sdjwtvc_storage.go
```diff
@@ -365,28 +368,42 @@ func getFirstCredentialInstanceFromBucket(bucket *bbolt.Bucket, aesKey [32]byte)
...
 	}
 
-	return &info, nil
+	instanceCount, err := getCredentialInstanceCount(bucket)
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to get instance count: %v", err)
+	}
+
+	return &SdJwtVcBatchMetadata{
```
kamphuisem, line 390: Why create a new struct here instead of only setting the `RemainingInstanceCount` property on the `info` object and returning a pointer to that instance?

irmaclient/sdjwtvc_storage.go
```diff
@@ -453,11 +441,11 @@ func (s *InMemorySdJwtVcStorage) RemoveCredentialByHash(hash string) ([]jwk.Key,
 	return nil, nil
 }
 
-func (s *InMemorySdJwtVcStorage) GetCredentialInfoList() irma.CredentialInfoList {
-	result := irma.CredentialInfoList{}
+func (s *InMemorySdJwtVcStorage) GetCredentialInfoList() []SdJwtVcBatchMetadata {
```
kamphuisem, line 444: Maybe add a comment as to why the function is named like this, because it might become kind of confusing later.

PR #490, opened 2025-09-01
Eudi Multiple Requestors

w-ensink:


server/conf.go
```diff
@@ -587,34 +653,46 @@ func (conf *Configuration) verifySdJwtIssuanceSettings() error {
...
+		if !ok {
+			return fmt.Errorf("expected cert chain for %v, but it's not present", issuerId)
+		}
+
+		x5cChain, err := utils.ConvertPemCertificateChainToX5cFormat(certChain)
+		if err != nil {
+			return fmt.Errorf("failed to convert cert chain for %v to x5c format: %v", issuerId, err)
+		}
+
+		issuerUrl, err := utils.ObtainIssuerUrlFromCertChain(certChain)
```
kamphuisem, line 684: Maybe good to log which `issuerUrl` is chosen to be configured, as certificates allow for multiple SAN-URI's.

server/conf.go
```diff
@@ -577,6 +581,68 @@ func (conf *Configuration) redisTLSConfig() (*tls.Config, error) {
 	return tlsConfig, nil
 }
 
+func readSdJwtIssuerPivKeys(dir string) (map[irma.IssuerIdentifier]*ecdsa.PrivateKey, error) {
+	entries, err := os.ReadDir(dir)
```
kamphuisem, line 585: Suggestion: use `filepath.Glob(filepath.Join(dir, "*.pem:"))` to filter the entries before looping over the entries.
That will make the check in the loop redundant.

server/conf.go
```diff
@@ -577,6 +581,68 @@ func (conf *Configuration) redisTLSConfig() (*tls.Config, error) {
...
+
+			result[issuerId] = privKey
+		}
+	}
+
+	return result, nil
+}
+
+func readSdJwtIssuerCertChains(dir string) (map[irma.IssuerIdentifier][]*x509.Certificate, error) {
+	entries, err := os.ReadDir(dir)
```
kamphuisem, line 616: Same suggestion: using `filepath.Glob(...)`

server/conf.go
```diff
@@ -587,34 +653,46 @@ func (conf *Configuration) verifySdJwtIssuanceSettings() error {
...
+	if len(privKeys) != len(certChains) {
+		return fmt.Errorf("the number of sdjwtvc cert chains and private keys don't match")
 	}
 
-	sdConf.X509CertChain = certChain
+	sdConf.Issuers = map[irma.IssuerIdentifier]*SdJwtIssuer{}
+	for issuerId, privKey := range privKeys {
+		certChain, ok := certChains[issuerId]
+		if !ok {
+			return fmt.Errorf("expected cert chain for %v, but it's not present", issuerId)
```
kamphuisem, line 676: Maybe only log the error and continue with the next key?
Would be kind of harsh to 'crash' the server on a mismatch in this configuration if you ask me.

PR #502, opened 2025-09-25
use prod trust anchors by default and allow for enabling staging ones

w-ensink:


eudi/eudiconfig.go
```diff
@@ -99,6 +101,48 @@ func (conf *Configuration) Reload() error {
...
+	)
+
+	conf.Verifiers.addRevocationListDistributionPoints(
+		Production_Yivi_RootCertificateRevocationListDistributionPoint,
+		Production_Yivi_VerifierCaCertificateRevocationListDistributionPoint,
+	)
+
+	// Read the hardcoded trust anchors
+	if err := conf.Issuers.addTrustAnchors([]byte(Production_Yivi_IssuerTrustAnchor)); err != nil {
+		return fmt.Errorf("failed to add yivi production staging issuer trust anchors: %v", err)
```
kamphuisem, line 117:
```suggestion
		return fmt.Errorf("failed to add yivi production issuer trust anchors: %v", err)
```

irmaclient/client.go
```diff
@@ -559,6 +565,9 @@ func (client *Client) rawLogEntriesToLogInfo(entries []*LogEntry) ([]LogInfo, er
 
 func (client *Client) SetPreferences(prefs Preferences) {
 	client.irmaClient.SetPreferences(prefs)
+	if prefs.DeveloperMode {
+		client.openid4vpClient.eudiConf.EnableStagingTrustAnchors()
```
kamphuisem, line 569: Maybe an idea to immediately update the revocation lists when activating staging anchors?
Otherwise, it might take up to half an hour for the lists to update themselfs.

PR #525, opened 2026-03-02
Fix VerifyMXRecord to handle implicit MX per RFC 5321

### LINKED ISSUE #524: VerifyMXRecord incorrectly rejects domains with A/AAAA records but no MX records
## Summary

The `VerifyMXRecord` function in `server/keyshare/email.go` incorrectly rejects email addresses for domains that have valid A/AAAA records but no MX records.

## Problem

Per RFC 5321 Section 5.1, when no MX records exist for a domain, mail servers should apply the "implicit MX" rule:

> "The lookup first attempts to locate an MX record associated with the
   name.  If a CNAME record is found, the resulting name is processed as
   if it were the initial name.  If a non-existent domain error is
   returned, this situation MUST be reported as an error.  If a
   temporary error is returned, the message MUST be queued and retried
   later (see [Section 4.5.4.1](https://www.rfc-editor.org/rfc/rfc5321.html#section-4.5.4.1)).  If an empty list of MXs is returned,
   the address is treated as if it was associated with an implicit MX
   RR, with a preference of 0, pointing to that host.  If MX records are
   present, but none of them are usable, or the implicit MX is unusable,
   this situation MUST be reported as an error."

https://www.rfc-editor.org/rfc/rfc5321.html#section-5.1

This means mail servers fall back to A/AAAA records of the domain itself. The current implementation attempts this fallback but has a bug in the control flow.

**Current behavior in `server/keyshare/email.go:167-204`:**

1. Looks up MX records
2. If no MX records found, checks for A/AAAA records
3. If A/AAAA records exist, continues (but doesn't return success)
4. Iterates over the MX records slice to validate hosts
5. Since MX records slice is empty, `hasValidHost` remains `false`
6. Returns `ErrInvalidEmailDomain`

**Expected behavior:**

When no MX records exist but valid A/AAAA records are found, the function should return `nil` (success).

## Steps to Reproduce

```go
err := VerifyMXRecord("user@sample.nl")
// Returns ErrInvalidEmailDomain even though sample.nl has a valid A record
```

```bash
$ dig MX sample.nl +short
# (no output - no MX records)

$ dig A sample.nl +short
1.1.1.1
```

## Suggested Fix

Return early after confirming A/AAAA records exist:

```go
if err != nil || len(records) == 0 {
    if records, err := net.LookupIP(host); err != nil || len(records) == 0 {
        return ErrInvalidEmailDomain
    }
    return nil  // A/AAAA fallback succeeded (implicit MX per RFC 5321 Section 5.1)
}
```

DibranMulder:
## Summary

- Fix `VerifyMXRecord` to correctly handle domains with A/AAAA records but no MX records
- Refactor DNS lookups to use interface-based dependency injection for testability

## Problem

Per RFC 5321 Section 5.1, when no MX records exist for a domain, mail servers should apply the "implicit MX" rule - treating the domain as if it had an MX record pointing to itself and using A/AAAA records for delivery.

The previous implementation checked for A/AAAA records as a fallback but didn't return success, causing valid email domains to be incorrectly rejected.

Fixes #524

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Fix
+- Bug that keyshare regisration failed when users email domain had no MX records.
 
 ## [0.19.2] - 2026-02-26
 ### Fix
 - Bug that caused HTTP request body to not be sent upon retransmission
 
-## Changed
+### Changed
 - Add wildcard support for authorized credentials and attributes in relying party and attestation provider certificates
 
 ## [0.19.1] - 2025-10-13
--- a/server/keyshare/email.go
+++ b/server/keyshare/email.go
@@ -15,6 +15,26 @@ import (
 	"github.com/privacybydesign/irmago/server"
 )
 
+// DNSResolver is an interface for DNS lookups, allowing injection of test doubles.
+type DNSResolver interface {
+	LookupMX(host string) ([]*net.MX, error)
+	LookupIP(host string) ([]net.IP, error)
+}
+
+// netDNSResolver is the default implementation using the net package.
+type netDNSResolver struct{}
+
+func (netDNSResolver) LookupMX(host string) ([]*net.MX, error) {
+	return net.LookupMX(host)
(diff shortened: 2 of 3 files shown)


w-ensink:
Looks good, but should be added to the changelog

w-ensink (approved):

rubenhensen (approved):

PR #529, opened 2026-03-26
Add local OpenID4VCI server and some integration tests

w-ensink:


eudi/didweb/document_resolver_test.go
```diff
@@ -63,6 +63,34 @@ func Test_didWebToURL(t *testing.T) {
 	}
 }
 
+func Test_Resolve_AllowInsecure_FallsBackToHTTP(t *testing.T) {
+	doc := did.Document{
+		Context: []string{"https://www.w3.org/ns/did/v1"},
+		ID:      "did:web:example.com",
+	}
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
kamphuisem, line 72: I'd check if you can use NewTlsServer instead. This way you can remove 'AllowInsecure' from the logic.

PR #532, opened 2026-04-02
OpenID4VP Compatibility

w-ensink:
- Add end-to-end OpenID4VP disclosure support for EUDI-issued SD-JWT-VC credentials (Veramo ecosystem), including key binding, presentation creation, and verifier validation
  - Implement claim_sets support in the EUDI DCQL handler, closing a spec compliance gap where disjunctive claim requirements were ignored
  - Add comprehensive Veramo integration test suite (20 tests) covering single/multi-credential disclosure, credential sets, optional credentials, value constraints, nested claims, arrays, non-SD claims, cross-ecosystem error handling, multiple vct_values, claim_sets, and an eduID credential type matching the real https://issuer.dev.eduid.nl/vct/eduid spec
  - Split verifier validation into composable x509_verifier_validator and did_verifier_validator with a composite dispatcher
  - Add SD-JWT presentation creation (CreatePresentation) with selective disclosure filtering
  - Fix linter issues (unused code, receiver name consistency, unnecessary fmt.Sprintf) and exclude testdata JSON from misspell checks in CI
  - Remove TranslatedString from AttributeValue (replaced by String)


eudi/openid4vp/eudi_sdjwt_dcql/handler.go
```diff
@@ -0,0 +1,281 @@
...
+			ExpiryDate:                  expiryUnix(batch),
+		})
+	}
+
+	return result, nil
+}
+
+// findBatches returns credential batches matching the query, with metadata
+// preloaded (including claim display names). When vct_values are specified,
+// only matching batches are returned; otherwise all batches are returned.
```
w-ensink: comment is wrong: no batches are returned when there's no match.

.github/workflows/status-checks.yml
```diff
@@ -105,6 +105,9 @@ jobs:
       - uses: actions/checkout@v4
       - name: Run all unit tests
         run: docker compose run test -v ./...
+      - name: Dump veramo container logs
+        if: failure()
```
w-ensink: this can be removed

eudi/openid4vp/did_verifier_validator.go
```diff
@@ -0,0 +1,201 @@
...
+	pubKey, didString, err := v.resolvePublicKey(clientId, preToken.Header)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to resolve verifier public key: %v", err)
+	}
+
+	// Parse and verify the JWT with the resolved key
+	var authRequest AuthorizationRequest
+	_, err = jwt.ParseWithClaims(requestJwt, &authRequest, func(token *jwt.Token) (any, error) {
+		typ, ok := token.Header["typ"]
+		if !ok {
```
kamphuisem, line 68: Merge `if` statements

eudi/openid4vp/did_verifier_validator.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// Determine a human-readable display name for the verifier. Priority:
+	// 1. response_uri hostname (consistent with regular IRMA sessions)
+	// 2. client_name from client_metadata (RFC 7591, best-effort)
+	// 3. domain from did:web
+	// 4. "unknown" (raw did:jwk is never useful to a user)
+	displayName := "unknown"
+	if host := hostFromURL(authRequest.ResponseUri); host != "" {
+		displayName = host
+	} else if authRequest.ClientMetadata.ClientName != "" {
```
kamphuisem: I would prefer this friendly name over de hostname (as that's exactly what it's for)

eudi/openid4vp/eudi_sdjwt_dcql/keybinding_adapter.go
kamphuisem, line 1: Why is this an adapter?
Shouldn't this be part of the `holderBindingKeyService`?

eudi/credentials/sdjwtvc/path_pointer.go
```diff
@@ -50,6 +55,17 @@ func (p *ProcessedSdJwtPayload) GetClaimValue(pathPointer []any) (any, error) {
 				return nil, fmt.Errorf("path component %d (%d): index out of range (length %d)", i, key, rv.Len())
 			}
 			current = rv.Index(key).Interface()
+		case float64:
```
kamphuisem, line 58: This change looks like a merge issue.
What does the type-checking for the array do here?

eudi/openid4vp/dcql/dcql.go
```diff
@@ -49,7 +49,7 @@ type CredentialQuery struct {
 	// OPTIONAL: An object defining additional properties requested by the verifier that apply to the
 	// metadata and validity data of the credential. The properties of this are defined per credential format.
 	// If omitted, no specific constraints are placed on the metadata or validity of the requested credential.
-	Meta Meta `json:"meta,omitempty"`
+	Meta Meta `json:"meta"`
```
kamphuisem: If it's optional, shouldn't it be a pointer?

eudi/openid4vp/dcql/dcql.go
```diff
@@ -83,19 +83,35 @@ type QueryResponse struct {
...
-//   - To address a particular claim within an object, append the key (claim name) to the array.
-//   - To address an element within an array, append the index to the array (as a non-negative, 0-based integer).
-//   - To address all elements within an array, append a null value to the array.
-type ClaimsPathPointer []string
+// JSON encoding: ["key", 0, null] → []any{"key", 0, nil}
+type ClaimsPathPointer []any
+
+// StringParts returns only the string components of the path, useful for
+// extracting the object key hierarchy (ignoring array indices and null).
+func (p ClaimsPathPointer) StringParts() []string {
```
kamphuisem: Why skip the other parts?
Aren't they important?

eudi/openid4vci/session.go
```diff
@@ -482,12 +481,11 @@ func (s *session) requestCredential(credentialConfigurationId string, cNonce *st
 	if err != nil {
 		// Error storing credentials; remove the already stored keys for the credentials that were received, to avoid orphaned keys in storage
 		if requireCryptographicKeyBinding {
-			// List all internal IDs of the keys that were stored for this credential request, to be able to remove them from storage
 			keyIds := make([]datatypes.UUID, len(publicKeyIdentifiers))
 			for i, key := range publicKeyIdentifiers {
 				keyIds[i] = key.ID
 			}
-			err := keyBindingService.RemoveKeys(keyIds)
+			err := services.NewHolderBindingKeyService(s.storage).RemoveKeys(keyIds)
```
kamphuisem: Now, 2 instances are created, which is not memory friendly.

eudi/openid4vp/eudi_sdjwt_dcql/handler.go
```diff
@@ -0,0 +1,408 @@
...
+		}
+
+		rawSdJwt := sdjwtvc.SdJwtVc(instance.RawCredential)
+
+		selected, err := sdjwtvc.CreatePresentation(rawSdJwt, sel.ClaimPaths)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create presentation: %w", err)
+		}
+
+		kbjwt, err := sdjwtvc.CreateKbJwt(selected, h.keyBinder, nonce, clientId)
```
kamphuisem: This will always generate a KB, which isn't always required.
We need to check the metadata to see if a KB is required or not.

eudi/openid4vp/irma_sdjwt_dcql/handler.go
```diff
@@ -104,7 +119,15 @@ func (h *SdJwtVcDcqlHandler) PrepareDisclosure(selections []dcql.DisclosureSelec
 			Credentials: []string{string(sdjwtWithKb)},
 		})
 
-		credLog := h.buildLogCredential(cred.Metadata, sel.AttributeNames)
+		// Extract flat attribute names from claim paths for log display.
```
kamphuisem: Shouldn't we retrieve the attribute names via the metadata?

eudi/openid4vp/did_verifier_validator.go
```diff
@@ -0,0 +1,201 @@
...
+		displayName = authRequest.ClientMetadata.ClientName
+	} else if domain, ok := didWebDomain(didString); ok {
+		displayName = domain
+	}
+
+	requestorInfo := &scheme.RelyingPartyRequestor{}
+	requestorInfo.Organization.LegalName = map[string]string{"en": displayName}
```
kamphuisem, line 95: The `client_name` could be a localized string, which we might want/need here ([see docs](https://www.rfc-editor.org/rfc/rfc7591.html#section-2.2)).

eudi/openid4vp/did_verifier_validator.go
```diff
@@ -0,0 +1,201 @@
...
+		didWeb := strings.TrimPrefix(clientId, "decentralized_identifier:")
+		return v.resolveDidWeb(didWeb, header)
+
+	default:
+		return nil, "", fmt.Errorf("unsupported client_id scheme: %s", clientId)
+	}
+}
+
+// resolveDidJwk extracts the public key from a did:jwk DID.
+func (v *DidVerifierValidator) resolveDidJwk(didJwk string, header map[string]any) (any, string, error) {
```
kamphuisem, line 142: I would rather put this func in the `didjwk` package for better availability.
The `(v *DidVerifierValidator)` part is not needed here.

eudi/openid4vp/did_verifier_validator.go
```diff
@@ -0,0 +1,201 @@
...
+	if err := jwk.Export(key, &rawKey); err != nil {
+		return nil, "", fmt.Errorf("failed to export raw key from did:jwk: %v", err)
+	}
+
+	return rawKey, didJwk, nil
+}
+
+// resolveDidWeb resolves a did:web DID document and extracts the verification key.
+func (v *DidVerifierValidator) resolveDidWeb(didWeb string, header map[string]any) (any, string, error) {
+	doc, err := v.didWebResolver.Resolve(didWeb)
```
kamphuisem, line 158: I would rather add this func to the `DocumentResolver`.

eudi/services/credential_service.go
```diff
@@ -99,75 +98,20 @@ func (c *credentialService) GetCredentialMetadataList() ([]*clientmodels.Credent
...
-								Type:   clientmodels.AttributeType_String,
-								String: &elemStr,
-							}
-						}
+				attrValue := clientmodels.NewAttributeValue(claimValue)
+
+				// Use the last element of the claim path as the attribute ID (e.g., ["address", "city"] → "city")
+				attrId := ""
+				if len(claimPath) > 0 {
+					if last, ok := claimPath[len(claimPath)-1].(string); ok {
```
kamphuisem: The claimPath could also end with an index (i.e. `0`) or `nil`, which would not work in this case.

eudi/openid4vp/did_verifier_validator.go
w-ensink: That would reduce the grannularity of the error message right?

eudi/openid4vp/did_verifier_validator.go
w-ensink: Agreed

eudi/openid4vp/eudi_sdjwt_dcql/keybinding_adapter.go
```diff
@@ -0,0 +1,90 @@
...
+
+// eudiKeyBindingStorage implements sdjwtvc.KeyBindingStorage by reading holder
+// binding private keys from the eudi SQLite storage.
+type eudiKeyBindingStorage struct {
+	keyStore storage.HolderBindingKeyStore
+}
+
+var _ sdjwtvc.KeyBindingStorage = (*eudiKeyBindingStorage)(nil)
+
+func (s *eudiKeyBindingStorage) GetAndRemovePrivateKey(pubKey jwk.Key) (*ecdsa.PrivateKey, error) {
```
w-ensink: Move to keybinder service

eudi/openid4vp/eudi_sdjwt_dcql/keybinding_adapter.go
w-ensink: done

eudi/credentials/sdjwtvc/path_pointer.go
w-ensink: When the claim path component is a number it should be used to index an array. Therefore we need to check that the claim it indexes into is an array.

eudi/openid4vp/did_verifier_validator.go
w-ensink: Theoretically yes, but as far as I know it's not even part of the Authorization request according to the openid4vp standard, so I'm not sure if we should even rely on it at all...

eudi/openid4vp/dcql/dcql.go
w-ensink: Done

eudi/openid4vp/dcql/dcql.go
w-ensink: It was only used in the irma_sdjwt part, there claim paths could only be a single string. But I removed it to make it less confusing.

eudi/openid4vci/session.go
w-ensink: Good point, fixed

eudi/openid4vp/eudi_sdjwt_dcql/handler.go
w-ensink: I added support for `require_crypthographic_holder_binding` flag in the credential query now.

eudi/services/credential_service.go
w-ensink: This is issuer metadata, the claim paths never contain integers or null values there, so this should not be an issue. I'll add a comment to clarify this.

eudi/openid4vp/irma_sdjwt_dcql/handler.go
w-ensink: Probably, but we'll make that correct when we implement the logging

internal/sessiontest/openid4vp_veramo_disclosure_test.go
```diff
@@ -0,0 +1,2755 @@
...
+							{
+								Path:        []any{"email"},
+								DisplayName: &clientmodels.TranslatedString{"en": "Email"},
+								Value:       strVal("dave@example.com"),
+							},
+						},
+					},
+				},
+			},
+			{Optional: true}, // optional phone (may have no owned options)
```
w-ensink: Look at this check more closely... should contain more detail

internal/sessiontest/openid4vp_veramo_disclosure_test.go
```diff
@@ -0,0 +1,2755 @@
...
+						Attributes: []expectedAttr{
+							{
+								Path:        []any{"email"},
+								DisplayName: &clientmodels.TranslatedString{"en": "Email"},
+								Value:       strVal("eve@example.com"),
+							},
+							{
+								Path:        []any{"domain"},
+								DisplayName: &clientmodels.TranslatedString{"en": "Domain"},
+								Value:       strVal("example.com"),
```
w-ensink: This one should have an requested value set

internal/sessiontest/openid4vp_veramo_disclosure_test.go
```diff
@@ -0,0 +1,2755 @@
...
+	}`
+	veramoSession := createVeramoVerifierDcqlSessionWithQuery(t, dcqlQuery)
+
+	startOpenID4VPDisclosureSession(t, c, veramoSession.RequestUri)
+
+	session := awaitSessionState(t, sessionHandler)
+	requireSessionState(t, session, 2, clientmodels.Type_Disclosure, clientmodels.Status_RequestPermission)
+
+	// Arrays are expanded into individual elements with indexed paths.
+	cred := session.DisclosurePlan.DisclosureChoicesOverview[0].OwnedOptions[0]
```
w-ensink: should compare the whole disclosure plan here

internal/sessiontest/openid4vp_veramo_disclosure_test.go
```diff
@@ -0,0 +1,2755 @@
...
+	// Path ["courses", 1] resolves to a specific element ("Databases"), not the whole array.
+	requireDisclosurePlan(t, session.DisclosurePlan, expectedDisclosurePlan{
+		Choices: []expectedPickOneChoice{
+			{
+				Owned: []expectedPlanCredential{
+					{
+						Attributes: []expectedAttr{
+							{
+								Path:  []any{"courses", 1},
+								Value: strVal("Databases"),
```
w-ensink, line 849: missing the display name

internal/sessiontest/openid4vp_veramo_disclosure_test.go
```diff
@@ -0,0 +1,2755 @@
...
+	}`
+	veramoSession := createVeramoVerifierDcqlSessionWithQuery(t, dcqlQuery)
+
+	startOpenID4VPDisclosureSession(t, c, veramoSession.RequestUri)
+
+	session := awaitSessionState(t, sessionHandler)
+	requireSessionState(t, session, 2, clientmodels.Type_Disclosure, clientmodels.Status_RequestPermission)
+
+	// Null path expands into individual elements with indexed paths.
+	cred := session.DisclosurePlan.DisclosureChoicesOverview[0].OwnedOptions[0]
```
w-ensink: should use the disclosure plan check

internal/sessiontest/openid4vp_veramo_disclosure_test.go
```diff
@@ -0,0 +1,2755 @@
...
+						Attributes: []expectedAttr{
+							{
+								Path:        []any{"given_name"},
+								DisplayName: &clientmodels.TranslatedString{"en": "Given name", "nl": "Voornaam"},
+								Value:       strVal("Student"),
+							},
+							{
+								Path:        []any{"is_student"},
+								DisplayName: &clientmodels.TranslatedString{"en": "IsStudent", "nl": "IsStudent"},
+								Value:       boolVal(true),
```
w-ensink: should have requested value present here

internal/sessiontest/openid4vp_veramo_disclosure_test.go
```diff
@@ -0,0 +1,2755 @@
...
+	veramoSession := createVeramoVerifierDcqlSessionWithQuery(t, dcqlQuery)
+
+	startOpenID4VPDisclosureSession(t, c, veramoSession.RequestUri)
+
+	session := awaitSessionState(t, sessionHandler)
+	requireSessionState(t, session, 3, clientmodels.Type_Disclosure, clientmodels.Status_RequestPermission)
+
+	plan := session.DisclosurePlan
+	require.NotNil(t, plan)
+	require.Len(t, plan.DisclosureChoicesOverview, 1)
```
w-ensink: use the requireDisclosure plan function

internal/sessiontest/openid4vp_veramo_disclosure_test.go
```diff
@@ -0,0 +1,2755 @@
...
+	})
+
+	cred := session.DisclosurePlan.DisclosureChoicesOverview[0].OwnedOptions[0]
+	grantPermission(t, c, session.Id, makeDisclosureChoice(cred))
+
+	session = awaitSessionState(t, sessionHandler)
+	requireSessionState(t, session, 2, clientmodels.Type_Disclosure, clientmodels.Status_Success)
+
+	result := checkVeramoVerifierOfferStatus(t, veramoSession.State)
+	require.Contains(t, []string{"VERIFIED", "RESPONSE_RECEIVED"}, result.Status)
```
w-ensink, line 1907: Add a check to verify the correct attributes/values were sent

internal/sessiontest/openid4vp_veramo_disclosure_test.go
```diff
@@ -0,0 +1,2755 @@
...
+	})
+
+	cred := plan.DisclosureChoicesOverview[0].OwnedOptions[0]
+	grantPermission(t, c, session.Id, makeDisclosureChoice(cred))
+
+	session = awaitSessionState(t, sessionHandler)
+	requireSessionState(t, session, 2, clientmodels.Type_Disclosure, clientmodels.Status_Success)
+
+	// The verifier may or may not verify the response (depends on verifier config),
+	// but the session should complete successfully without a KB-JWT.
```
w-ensink: how does this verify no kbjwt was included?

internal/sessiontest/openid4vp_veramo_disclosure_test.go
```diff
@@ -0,0 +1,2755 @@
...
+		claim([]any{"email"}, "batch2@example.com"),
+	)
+
+	// Step 4: Third disclosure — all instances exhausted, should fail.
+	veramoSession3 := createVeramoVerifierDcqlSessionWithQuery(t, dcqlQuery)
+	startOpenID4VPDisclosureSession(t, c, veramoSession3.RequestUri)
+
+	session = awaitSessionState(t, sessionHandler)
+	require.Equal(t, clientmodels.Status_Error, session.Status,
+		"third disclosure should fail because all batch instances are exhausted")
```
w-ensink, line 2381: should a lack of instances be handled by giving a session error or by providing owned options that are expired?

internal/sessiontest/openid4vp_veramo_disclosure_test.go
```diff
@@ -0,0 +1,2755 @@
...
+
+	startOpenID4VPDisclosureSession(t, c, veramoSession.RequestUri)
+
+	session := awaitSessionState(t, sessionHandler)
+
+	// The IRMA handler finds the credential matching test.test.email and offers it.
+	// The session reaches permission request with the IRMA credential as an option.
+	require.Equal(t, clientmodels.Status_RequestPermission, session.Status)
+	require.NotNil(t, session.DisclosurePlan)
+	require.NotEmpty(t, session.DisclosurePlan.DisclosureChoicesOverview)
```
w-ensink: use requireDisclosurePlan

internal/sessiontest/openid4vp_veramo_disclosure_test.go
w-ensink: it can't get the result from the verifier server, so we'll just rely on the verifier to reject an sdjwt with kbjwt in this case...

PR #536, opened 2026-04-21
Add logs for OpenID clients & SD-JWT removal

w-ensink:
## Summary

This PR introduces a dedicated EUDI activity logging system and unifies image handling across the client.

**EUDI log service** — A new `EudiLogService` backed by SQLCipher stores activity logs for OpenID4VCI issuance, OpenID4VP disclosure, and EUDI credential removal. Previously, OpenID4VP disclosure logs were forced into the IRMA bbolt store via a conversion layer; they now live in their own tables (`eudi_log_entries` / `eudi_log_credentials`) with proper schema.

**Merged log timeline** — `LoadNewestLogs` and `LoadLogsBefore` now query both bbolt (IRMA) and SQLCipher (EUDI) and merge results by timestamp. Pagination changed from index-based to time-based to work across both backends.

**`ImagePath` → `Image` migration** — All `ImagePath *string` fields on `TrustedParty`, `Credential`, `CredentialDescriptor`, `SelectableCredentialInstance`, and `LogCredential` are replaced with inline `Image *Image` (base64-encoded). A new `ImageFromFile()` helper reads logos from disk at the point of use. This eliminates the iOS logo-path bug class entirely — logos survive storage moves without path re-resolution.

**Hash deduplication** — `hashForSdJwtVc` now strips standard JWT claims (`iss`, `iat`, `exp`, `nbf`, `sub`, `vct`, `cnf`, `status`, `_sd`, `_sd_alg`, `fed`) before hashing, so re-issuing the same credential with different timestamps produces the same hash and deduplicates into one batch.

**Other changes:**
- `DeleteBatch` now cascades to holder binding keys
- `RemoveStorage()` clears EUDI database rows and filesystem contents (logos, certificates, CRLs)
- `StandardClaims` set centralized in `sdjwtvc` package (was duplicated in multiple places)
- Issuer display name and logo now included in EUDI disclosure logs
- Issuer display metadata added to the test issuer configuration

## Breaking changes

- `LoadLogsBefore(beforeIndex uint64, max int)` → `LoadLogsBefore(before time.Time, max int)`
- `LogInfo.ID` field removed
- `ImagePath` fields removed from all client model types
- `openid4vci.Handler.Success` signature changed: `Success(result string)` → `Success(result string, issuedCredentials []*clientmodels.Credential)`

## Test plan

- [x] Unit tests for `EudiLogService` round-trips (issuance, disclosure, removal, pagination)
- [x] Unit tests for `hashForSdJwtVc` (determinism, key-order independence, metadata stripping)
- [x] Unit test for cascading deletion of holder binding keys
- [x] Integration tests for OID4VCI issuance logs (pre-auth & auth-code flows)
- [x] Integration test for OID4VCI denied permission creates no log
- [x] Integration tests for OID4VP disclosure logs (including issuer name + image)
- [x] Integration test for EUDI credential removal logs (with attributes)
- [x] Integration test for deeply nested credential logs (issuance + removal)
- [x] Integration test for selective disclosure log (only shared attributes appear)
- [x] Integration test for duplicate credential deduplication + removal
- [x] Integration test for cross-backend chronological merge (IRMA + EUDI)
- [x] Integration test for `LoadLogsBefore` pagination across both backends
- [x] Integration test for `RemoveStorage` clearing EUDI database + filesystem
- [x] Integration test for credential store items having images
- [x] iOS logo path bug tests updated to verify base64 images survive storage moves

🤖 Generated with [Claude Code](https://claude.com/claude-code)


client/client.go
```diff
@@ -372,8 +373,35 @@ func (client *Client) RemoveCredentialsByHash(hashByFormat map[clientmodels.Cred
...
 
-	// Delete EUDI credentials.
+	// Delete EUDI credentials (read metadata first for the removal log).
 	if len(eudiHashes) > 0 {
+		credentialService := services.NewCredentialService(client.eudiStorage)
+		allEudiCreds, err := credentialService.GetCredentialMetadataList()
+		if err != nil {
+			return fmt.Errorf("failed to read eudi credentials for removal log: %v", err)
+		}
+
```
w-ensink, line 383: unrelated, but does this function also delete holder binding private keys?

common/clientmodels/types.go
```diff
@@ -31,6 +33,19 @@ type Image struct {
 	MimeType *string `json:"mime_type,omitempty"`
 }
 
+// ImageFromFile reads an image file from disk and returns it as a base64-encoded Image.
+// Returns nil if the path is empty or the file cannot be read.
+func ImageFromFile(path string) *Image {
```
w-ensink, line 38: should this function be here...?

eudi/openid4vci/handler.go
```diff
@@ -25,7 +25,7 @@ type SessionDismisser interface {
 
 type Handler interface {
 	// Shared interface functions with irmaclient.Handler
-	Success(result string)
+	Success(result string, issuedCredentials []*clientmodels.Credential)
```
w-ensink, line 28: why does this need the issued credentials as a parameter?

eudi/openid4vp/eudi_sdjwt_dcql/handler.go
```diff
@@ -96,22 +96,24 @@ func (h *SdJwtVcDcqlHandler) FindCandidates(query dcql.CredentialQuery) (*dcql.C
...
-		credentialLogoManager := h.storage.FileSystem().Credentials().LogoManager()
-		for _, display := range batch.CredentialMetadata.Display {
-			if display.LogoURI != "" {
-				filename := credentialLogoManager.GetLogoFilenameWithoutExtensionFromUrl(display.LogoURI)
-				imageData, err := credentialLogoManager.GetLogo(filename)
-				if err != nil {
-					eudi.Logger.Warnf("failed to get credential logo from %q: %v", display.LogoURI, err)
-					continue
-				}
+		if batch.CredentialMetadata != nil {
```
w-ensink: in what scenario whould credential metadata be nil?

eudi/services/credential_service.go
```diff
@@ -559,16 +568,45 @@ func lookupDisplayName(lookup map[string]clientmodels.TranslatedString, path []a
 	return nil, false
 }
 
+// issuerMetadataKeys lists JWT-registered and SD-JWT-specific claims that must be
```
w-ensink: Can this be `const` and reused?

eudi/services/log_service.go
```diff
@@ -0,0 +1,285 @@
...
+				formats = append(formats, f)
+			}
+		}
+		result[i] = models.EudiLogCredential{
+			ID:           datatypes.NewUUIDv4(),
+			CredentialId: c.CredentialId,
+			Formats:      mustJSON(formats),
+			Name:         mustJSON(c.Name),
+			IssuerName:   mustJSON(c.Issuer.Name),
+			Attributes:   mustJSON(c.Attributes),
```
w-ensink: what is mustJson?

eudi/services/log_service.go
```diff
@@ -0,0 +1,285 @@
...
+			if err == nil && imageData != nil {
+				image = &clientmodels.Image{Base64: *imageData}
+			}
+		}
+		result[i] = clientmodels.LogCredential{
+			CredentialId: c.CredentialId,
+			Formats:      formats,
+			Name:         name,
+			Image:        image,
+			Issuer:       clientmodels.TrustedParty{Name: issuerName},
```
w-ensink: should trusted party also contain a logo?

eudi/services/credential_service.go
w-ensink: no, it's not primitive. Now it's shared across all use cases tho

eudi/openid4vp/eudi_sdjwt_dcql/handler.go
w-ensink: In case no metadata is available for whatever reason...

eudi/openid4vci/handler.go
w-ensink: It's used for the logs. This completely decouples the openid4vci client from the logging system

client/client.go
w-ensink: It does now.

client/client.go
```diff
@@ -443,26 +470,79 @@ func (client *Client) RemoveStorage() error {
...
+			i++
+		default:
+			merged = append(merged, b[j])
+			j++
+		}
+	}
+	return merged
 }
 
 func (client *Client) rawLogEntryToLogInfo(entry *irmaclient.LogEntry) (clientmodels.LogInfo, error) {
```
kamphuisem: Should we move this to irma_adapters.go?
I think it would be good to split the conversion/adapter-like functions per type, so it doesn't clutter the client.go file to much.

client/openid4vci_adapters.go
```diff
@@ -33,8 +34,21 @@ func (a *openid4vciSessionAdapter) Cancelled() {
...
+
+	// Store issuance log.
+	if len(issuedCredentials) > 0 {
+		logService := services.NewEudiLogService(a.session.client.eudiStorage)
+		if err := logService.AddIssuanceLog(
+			clientmodels.Protocol_OpenID4VCI,
+			a.session.State.Requestor,
+			issuedCredentials,
+		); err != nil {
+			irma.Logger.Errorf("failed to store openid4vci issuance log: %v", err)
```
kamphuisem: There should also be an `eudi.Logger` which you can use.
This will reduce dependency on `irma`

client/openid4vp_adapters.go
```diff
@@ -30,11 +27,11 @@ func (a *openid4vpSessionAdapter) Cancelled() {
...
-	// Store the disclosure log
-	if a.session.client.logsStorage != nil && len(credentialLogs) > 0 {
-		logEntry := openid4vpCredentialLogsToIrmaclientLogEntry(credentialLogs, a.session.State.Requestor)
-		if err := a.session.client.logsStorage.AddLogEntry(logEntry); err != nil {
-			irma.Logger.Errorf("failed to store openid4vp log: %v", err)
+	// Store the disclosure log in the EUDI SQLCipher database.
+	if len(credentialLogs) > 0 {
+		logService := services.NewEudiLogService(a.session.client.eudiStorage)
+		if err := logService.AddDisclosureLog(a.session.State.Requestor, credentialLogs); err != nil {
+			irma.Logger.Errorf("failed to store openid4vp disclosure log: %v", err)
```
kamphuisem: Use `eudi.Logger`

eudi/credentials/sdjwtvc/sdjwtvc.go
```diff
@@ -219,12 +219,29 @@ const (
...
+	Key_Federation               string = "fed"
 
 	SdJwtVcTyp        string = "dc+sd-jwt"
 	SdJwtVcTyp_Legacy string = "vc+sd-jwt"
 	KbJwtTyp          string = "kb+jwt"
 )
 
+// StandardClaims contains JWT-registered and SD-JWT-specific claims that are not user data.
+// Use this to distinguish issuer/protocol metadata from actual credential attributes.
+var StandardClaims = map[string]struct{}{
```
kamphuisem, line 231: Why is this a `map[string]struct{}` instead of just a array/slice?

eudi/services/log_service.go
```diff
@@ -0,0 +1,375 @@
...
+func NewEudiLogService(s storage.Storage) EudiLogService {
+	return &eudiLogService{
+		store:               db.NewEudiLogStore(s.Db()),
+		credLogoManager:     s.FileSystem().Credentials().LogoManager(),
+		issuerLogoManager:   s.FileSystem().Issuers().LogoManager(),
+		verifierLogoManager: s.FileSystem().Verifiers().LogoManager(),
+	}
+}
+
+func (s *eudiLogService) AddIssuanceLog(protocol clientmodels.Protocol, issuer clientmodels.TrustedParty, credentials []*clientmodels.Credential) error {
```
kamphuisem: `AddIssuanceLog` and `AddDisclosureLog` have virtually the same implementation.
Maybe good to extract this to a seperate func?

eudi/services/log_service.go
```diff
@@ -0,0 +1,375 @@
...
+	if _, err := manager.SaveLogo(filename, rawBytes); err != nil {
+		return ""
+	}
+	return filename
+}
+
+// saveRequestorLogo persists the requestor's logo image (if any) to the
+// verifier logo storage and returns the filename used for storage.
+// Returns "" if the requestor has no image.
+func (s *eudiLogService) saveRequestorLogo(tp clientmodels.TrustedParty) string {
```
kamphuisem: I would not create this func, as it is unclear from the functions name if the requestor should be an issuer or a verifier. I'd rather use the `saveLogoFromBase64` as it is used in the other places in this file.

eudi/storage/db/models/logs.go
```diff
@@ -0,0 +1,60 @@
...
+// (OpenID4VCI issuance, OpenID4VP disclosure, or credential removal).
+type EudiLogEntry struct {
+	ID        datatypes.UUID `gorm:"type:uuid;primaryKey"`
+	Type      string         // "issuance", "disclosure", "removal"
+	Protocol  string         // "openid4vci", "openid4vp", or empty for removal
+	CreatedAt time.Time      `gorm:"index"`
+
+	// Requestor/verifier/issuer info (JSON-encoded TranslatedString for name).
+	RequestorId           string
+	RequestorName         datatypes.JSON `gorm:"type:json"`
```
kamphuisem, line 19: Are all locale names stored? I.e. do we store en + nl + ... ?

eudi/storage/db/models/logs.go
```diff
@@ -0,0 +1,60 @@
...
+	Name       datatypes.JSON `gorm:"type:json"`
+	IssuerName datatypes.JSON `gorm:"type:json"`
+	IssuerId   string
+
+	// JSON-encoded []clientmodels.Attribute — full attribute list with paths,
+	// display names, and values. Stored as a blob since logs are write-once.
+	Attributes datatypes.JSON `gorm:"type:json"`
+
+	// Credential timing and status metadata.
+	IssuanceDate        int64
```
kamphuisem: `time.Time`?

eudi/storage/db/models/logs.go
```diff
@@ -0,0 +1,60 @@
...
+	ExpiryDate          int64
```
kamphuisem: `time.Time`?

eudi/storage/db/models/logs.go
```diff
@@ -0,0 +1,60 @@
...
+	RevocationSupported bool
+	IssueURL            datatypes.JSON `gorm:"type:json"` // JSON-encoded *TranslatedString, nil when not set.
+
+	// Filename (without extension) of the credential logo stored by the
+	// credential logo manager. Empty when no logo is available.
+	LogoFilename string
+
+	// Filename (without extension) of the issuer logo stored by the
+	// issuer logo manager. Empty when no logo is available.
+	IssuerLogoFilename string
```
kamphuisem, line 59: No option for verifier logo?

eudi/storage/db/credential_store.go
```diff
@@ -168,6 +172,16 @@ func (s *credentialStore) DeleteBatch(batchID datatypes.UUID) error {
 		return fmt.Errorf("batchID is required")
 	}
 
+	// Explicitly delete holder binding keys (and their cascaded metadata) that belong
+	// to this batch's instances. The FK lives on IssuedCredentialInstance (belongs-to),
+	// so the DB-level ON DELETE CASCADE does not cover this direction.
+	if err := s.db.Exec(`DELETE FROM holder_binding_keys WHERE id IN (
```
kamphuisem: Hmm, should we change the association to a [Has One](https://gorm.io/docs/has_one.html) type to make this work automatically?

eudi/storage/storage.go
```diff
@@ -93,11 +95,18 @@ func (s *storage) Close() error {
 }
 
 func (s *storage) RemoveAll() error {
-	return s.db.Transaction(func(tx *gorm.DB) error {
-		result := tx.Session(&gorm.Session{AllowGlobalUpdate: true}).
-			Delete(&models.HolderBindingKey{}). // CASCADE should take care of deleting related metadata
-			Delete(&models.CredentialBatch{})   // CASCADE should take care of deleting related instances and metadata
+	if err := s.db.Transaction(func(tx *gorm.DB) error {
+		session := tx.Session(&gorm.Session{AllowGlobalUpdate: true})
+		if err := session.Delete(&models.HolderBindingKey{}).Error; err != nil {
```
kamphuisem: This should be deleted on cascading down from deleting CredentialBatch

irma/server/requestorserver/server.go
```diff
@@ -77,16 +79,12 @@ func (s *Server) Start(config *Configuration) error {
...
 	var err error
 	for i := 0; i < cap(done); i++ {
 		if err = <-done; err != nil {
 			_ = server.LogError(err)
 		}
-		if !stopped {
-			stopped = true
-			close(s.stop)
-		}
+		s.stopOnce.Do(func() { close(s.stop) })
```
kamphuisem, line 87: What does this improve?
And is the `stop chan struct{}` still being used for something?

eudi/credentials/sdjwtvc/sdjwtvc.go
w-ensink: Makes lookup a bit simpler and I think it's more correct because it's a set of claims (there are no duplicates)

eudi/storage/db/models/logs.go
w-ensink: This is a log for a single credential. A credential doesn't have a verifier. The verifier is stored at the DisclosureLog level.

eudi/storage/db/models/logs.go
w-ensink: Good point, done

eudi/services/log_service.go
w-ensink: Good point, done

client/openid4vp_adapters.go
w-ensink: Good point, done

client/client.go
w-ensink: Good point, done

eudi/storage/db/credential_store.go
w-ensink: I think I've done that now, but maybe you can double check if this is what you mean?

irma/server/requestorserver/server.go
w-ensink: Some tests were flaky due to a race condition in the shutdown of this server. This change makes sure the shutdown only happens once.

eudi/services/log_service.go
w-ensink: Good point, done

eudi/storage/db/models/logs.go
w-ensink: Yes, we store all translations that have been downloaded by the VP/VCI client.

kamphuisem (approved):

PR #538, opened 2026-04-29
Add FS level file name hashing, salted with AES derived key

w-ensink:
## Move FS filename hashing into the FS layer; salt with AES-derived key

  Refactors the EUDI filesystem (`eudi/storage/filesystem/`) so all on-disk filenames are derived inside the FS layer using `HMAC-SHA256(fs_filename_key,
  logical_name)`, where `fs_filename_key` is HKDF-derived from the existing AES storage key. Plaintext URLs and IDs no longer land on disk via filenames;
  an attacker with read access to the storage directory but without the AES key cannot test "is logo for URL X cached?".

  Design doc: `docs/refactor-fs-filename-hashing.md` (full construction, threat model, alternatives considered, decisions).

  ### What changes

  **FS layer** (`eudi/storage/filesystem/`)
  - `NewFileSystemStorage(aesKey [32]byte, basePath string)` — takes the AES key directly. AES-GCM middleware and the HKDF-derived `fs_filename_key` (HKDF
  info: `"irmago-fs-filename-v1"`) are built internally.
  - New `scopedFS` type — bound to a directory; `Write`/`Read`/`Exists`/`Delete`/`RemoveAll`/`Walk` take logical names. HMAC + hex happens internally;
  encryption stays in the same path.
  - `LogoManager`: `Save(key, data) error`, `Get(key) []byte`, `Exists(key) bool`, `RemoveAll() error`. Returns raw bytes; base64 wrapping moved to the
  service layer.
  - `CertificateRevocationListManager`: `Save(crl, distPoint)`, `Read(distPoint)`, `Exists(distPoint)`, `Remove(distPoint)`, `LoadAll(onError
  func(error))`, `RemoveAll()`.
  - Deleted: `GetLogoFilenameWithoutExtensionFromUrl`, `GetCrlFileNameForCertDistributionPoint`, `NewStorageMiddleware`, `GetAllFileNames`,
  `ReadFromFileName`, `RemoveByFileName`, `CrlExists`.
  - Certificate manager (`{signature_hex}.pem`) intentionally untouched — its filename is the cert's public signature, not a hash of secret input; salting
  buys nothing.

  **DB schema**
  - Dropped `EudiLogEntry.RequestorLogoFilename`, `EudiLogCredential.LogoFilename`, `EudiLogCredential.IssuerLogoFilename`. Read path uses the adjacent ID
  columns (`RequestorId`, `CredentialId`, `IssuerId`) + `manager.Exists`/`manager.Get`. Greenfield — no migration (this code is not yet in production).

  **Call sites**
  - `client/client.go`: stops constructing the FS encryption middleware; the AES key flows directly into EUDI `storage.NewStorage`. The middleware is still
   constructed for the IRMA classic `clientstorage`.
  - `client/irma_adapters.go`, `eudi/services/credential_service.go`, `eudi/openid4vci/client.go` + `session.go`, `eudi/openid4vp/client.go` +
  `eudi_sdjwt_dcql/handler.go`: replaced the "hash URL → call manager" two-step with single logical-key calls. Base64 encoding done at
  `clientmodels.Image{Base64: ...}` construction.
  - `eudi/eudiconfig.go`: `ResolveVerifierLogo` returns `[]byte`.
  - `eudi/trustmodel.go`: distribution-point-based CRL operations throughout; `loadRevocationLists` uses `LoadAll(onError)`.
  - ~50 test sites that previously passed `&mocks.MockEncryptionMiddleware{}` to `NewFileSystemStorage` now pass `[32]byte{}` (real AES-GCM with a zero key
   for tests). Mock middleware itself is kept — IRMA classic still uses it.


eudi/services/log_service.go
```diff
@@ -294,19 +277,35 @@ func modelCredentialsToLogCredentials(creds []models.EudiLogCredential, credLogo
...
-		return ""
+	if err := manager.Save(key, rawBytes); err != nil {
+		eudi.Logger.Warnf("failed to cache logo for key %q: %v", key, err)
+	}
+}
+
+// loadLogoImage looks up a previously-saved logo by its logical key and wraps
+// it as a base64-encoded clientmodels.Image. Returns nil when the key is empty,
+// the manager is unavailable, or no logo is cached.
+func loadLogoImage(manager filesystem.LogoManager, key string) *clientmodels.Image {
```
w-ensink: Why is this not used in more places?

kamphuisem (approved):

PR #546, opened 2026-05-20
Add `urn:...` vct support

w-ensink:
Add `urn:...` vct value support and add eudi issuer to integration tests


w-ensink:
/dobby i fixed your comments, review again

kamphuisem (approved):

PR #556, opened 2026-06-04
Prefer SD-JWT VC metadata over VCI issuer metadata

w-ensink:
[OpenID4VCI spec 12.2.4](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-issuer-metadata-p) states:
> credential_metadata: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata. **Format-specific mechanisms, such as SD-JWT VC display metadata are always preferred by the Wallet over the information in this object, which serves as the default fallback**. 

Therefore we should prefer using SD-JWT VC metadata.

This PR also improves the SD-JWT VC metadata support itself with the type extension and integrity checks. It will merge the metadata with the VCI issuer metadata for a given credential, so that the VCI metadata is used for missing claim display names (on a locale basis).


eudi/openid4vci/client.go
```diff
@@ -484,3 +475,129 @@ func handleFailure(handler Handler, message string, fmtArgs ...any) {
...
+		}
+
+		resolved, err := resolver.Resolve(ctx, config.VerifiableCredentialType, client.allowInsecureHttp)
+		if err != nil {
+			eudi.Logger.Infof("vct type metadata resolution failed for %q (vct=%q): %v; falling back to credential_metadata", configID, config.VerifiableCredentialType, err)
+			continue
+		}
+
+		mapped := mapVctToCredentialMetadata(resolved)
+		config.CredentialMetadata = &mapped
```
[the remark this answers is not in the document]
w-ensink: fixed, now merged per locale

eudi/credentials/sdjwtvc/typemetadata/resolver.go
```diff
@@ -0,0 +1,212 @@
...
+//
+// vctURL must use https://; http:// is accepted only when devMode is true.
+// Returns an error on: scheme rejection, fetch failure, parse failure,
+// extends-chain depth overflow, extends-chain cycle, or extends#integrity
+// mismatch.
+//
+// The cache (raw bytes + parsed) is populated for every URL visited. Use
+// RawDocument to retrieve the raw bytes later (e.g. for post-issuance
+// integrity verification against the issued JWT's vct#integrity claim).
+func (r *Resolver) Resolve(ctx context.Context, vctURL string, devMode bool) (*VctTypeMetadata, error) {
```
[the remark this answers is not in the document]
w-ensink: This is visual metadata only so not super relevant for an attacker

eudi/credentials/sdjwtvc/typemetadata/resolver.go
```diff
@@ -0,0 +1,253 @@
...
+
+// DefaultMaxExtendsDepth caps the extends chain length to prevent runaway
+// fetching. The number is a safety stop; real chains are typically very short.
+const DefaultMaxExtendsDepth = 10
+
+// integrityHashers lists the hash algorithms accepted in <algo>-<base64>
+// integrity strings (vct#integrity, extends#integrity). The set mirrors
+// W3C SRI's three required algorithms; weaker algorithms (md5, sha1) are
+// intentionally absent. SD-JWT VC § 5 defers algorithm choice to SRI.
+var integrityHashers = map[string]func() hash.Hash{
```
kamphuisem, line 25: I think something similar is available in the `iana` internal package. Maybe check if you can re-use that.

eudi/credentials/sdjwtvc/typemetadata/typemetadata.go
```diff
@@ -31,14 +31,25 @@ type VctTypeMetadata struct {
...
+	// this document refines, per SD-JWT VC § 6.1.
+	Extends string
+	// ExtendsIntegrity is the integrity hash for the parent document, used to
+	// verify the document referenced by Extends. Format: "<algo>-<base64>".
+	ExtendsIntegrity string
 }
 
 // DisplayEntry is one localized display entry from the type-metadata document.
-// The SD-JWT VC spec uses "lang", not "locale".
+// Current SD-JWT VC drafts (draft-13 onward) use "locale"; earlier drafts used "lang".
```
kamphuisem, line 52: Maybe good to note which version we actually implement

eudi/openid4vci/merge_metadata.go
kamphuisem, line 1: Maybe `merge,go` isn't the best filename.
I'd rather name it `merge_metadata.go`.

Also, don't we need merge functionality in OpenID4VP?

eudi/openid4vci/session.go
```diff
@@ -193,6 +246,147 @@ func (s *session) obtainCredentials(accessToken string) ([]*fetchedCredential, e
...
+			}
+			break
+		}
+	}
+	return nil
+}
+
+// vctIntegrityClaim is the JWT claim name carrying the integrity hash of the
+// SD-JWT VC type metadata document referenced by the credential's vct field.
+const vctIntegrityClaim = "vct#integrity"
```
kamphuisem: Define this in `sdjwtvc` package.

eudi/openid4vci/session.go
```diff
@@ -193,6 +246,147 @@ func (s *session) obtainCredentials(accessToken string) ([]*fetchedCredential, e
...
+	}
+	return nil
+}
+
+// lookupVctIntegrityClaim returns the vct#integrity claim string if present
+// and a non-empty string. A claim present with a non-string (or empty) value
+// is an error: the JWT shape for vct#integrity is defined as a string, so a
+// non-string here is either an issuer bug or a downgrade attempt that
+// shouldn't silently bypass integrity verification.
+func lookupVctIntegrityClaim(payload sdjwtvc.ProcessedSdJwtPayload) (string, bool, error) {
```
kamphuisem: Move to `sdjwtvc` package

eudi/credentials/sdjwtvc/typemetadata/typemetadata.go
w-ensink: Hard to say, we have some parts from different versions of the specs, as we have to deal with with real life scenarios. We don't target a single version.

eudi/openid4vci/session.go
w-ensink: Good point, done

eudi/openid4vci/session.go
w-ensink: Good point, done

eudi/openid4vci/merge_metadata.go
w-ensink: I renamed it, but the metadata is persisted for OpenID4VP. For missing credentials it's not usable because there only SD-JWT metadata is available (if even that).

eudi/credentials/sdjwtvc/typemetadata/resolver.go
w-ensink: Can't be used unfortunately because the names are slightly different (`sha256` for `sri` vs `sha-256` for `iana`)

PR #561, opened 2026-06-08
Add status list support for SD-JWT

### LINKED ISSUE #516: OID4VCI - DIIPv4: Add support for IETF Token Status List
See [DIIPv4 specs, 4.7](https://fidescommunity.github.io/DIIP/v4#validity-and-revocation-algorithm)

The spec requires the use of a status list mechanism for validity/revocation checking, which we need to implement.

w-ensink:
Implementing [Token Status List (TSL) draft 15](https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-15.html). We don't implement the latest version of the spec yet, because [DIIP v5](https://fidescommunity.github.io/DIIP/) is specifically talking about this draft.

Fixes privacybydesign/irmamobile#516


w-ensink:
/dobby please give me your final review

client/client.go
```diff
@@ -211,6 +238,15 @@ func (client *Client) Close() error {
 	return client.storage.Close()
 }
 
+// RefreshStatuses re-fetches the Token Status List for every stored
+// SD-JWT VC instance and updates its LastKnownStatus column. Use
+// this on app resume or when the UI exposes an explicit refresh
+// action. Errors during the sweep are logged; the previous
+// LastKnownStatus persists for any URI that fails to refresh.
+func (client *Client) RefreshStatuses(ctx context.Context) error {
```
kamphuisem, line 249: Why expose this method publicly on Client?
I think it should be private, as we don't want other callers to be able to call this method repeatedly.

client/client.go
```diff
@@ -586,7 +620,7 @@ func (client *Client) GetPreferences() clientsettings.Preferences {
 	return client.irmaClient.Preferences
 }
 
-func (client *Client) InitJobs(eudiRevocationListUpdateInterval time.Duration) {
+func (client *Client) InitJobs(eudiRevocationListUpdateInterval, statusRefreshInterval time.Duration) {
```
kamphuisem: With a new revocation 'type', maybe good to rename the `eudiRevocationListUpdateInterval` to something like `eudiCrlUpdateInterval`, so we can more clearly tell the difference.
I'd name the new variable `statusTokenListRefreshInterval`, as `status` is too common.

w-ensink:
/dobby i changed the architecture a bit, please review it one more time

eudi/credentials/statuslist/context.go
```diff
@@ -0,0 +1,71 @@
...
+	FetchTimeout time.Duration
+}
+
+// payloadFromToken extracts the Status List Token's claims from a
+// signature-verified jwt.Token into our statusListPayload struct via a
+// JSON round-trip — the struct's json tags mirror the spec's claim
+// names (iat/exp marshal as Unix seconds, status_list as a nested
+// object). Returns an error if the mandatory status_list claim is
+// missing or a claim is shaped wrong for its struct field.
+func payloadFromToken(token jwt.Token) (statusListPayload, error) {
```
kamphuisem: Why is this in `context.go`?
There must be a better place for this, as it's got nothing to do with context.

eudi/credentials/statuslist/doc.go
kamphuisem, line 1: I'd rather place this comment somewhere in `checker.go` instead of an 'empty' file.

eudi/credentials/statuslist/reference.go
```diff
@@ -0,0 +1,98 @@
...
+	URI   string `json:"uri"`
+}
+
+// Status is the typed value returned by Checker.Check.
+//
+// Mapping from the raw bit-value n is:
+//
+//	n == 0 -> StatusValid
+//	n == 1 -> StatusInvalid
+//	n == 2 -> StatusSuspended  (requires bits >= 2)
```
kamphuisem, line 34: Not really a PR comment, but rather a question; what do we show our users for the `suspended` status?

eudi/credentials/statuslist/verifier.go
```diff
@@ -0,0 +1,148 @@
...
+// ClockSkewSeconds matches sdjwtvc.ClockSkewInSeconds. Kept local to
+// avoid an import cycle between statuslist and sdjwtvc.
+const ClockSkewSeconds = 180
+
+// statusListClaim mirrors the inner `status_list` object of a Status
+// List Token payload (draft-ietf-oauth-status-list-15 §6).
+type statusListClaim struct {
+	Bits int    `json:"bits"`
+	Lst  string `json:"lst"`
+	// AggregationURI is intentionally not parsed — v1 ignores aggregation.
```
kamphuisem: Maybe explicitly write that aggregation is mainly used by relying parties (otherwise, we will probably end up fetching status lists, which we have no credentials for anyway)

eudi/jwt/key_providers.go
kamphuisem, line 1: How is this different from the `SdJwtKeyProvider` but on another level?
Maybe check if we can combine the 2 (maybe in a seperate package or something).

eudi/openid4vp/eudi_sdjwt_dcql/handler.go
```diff
@@ -46,6 +46,13 @@ func isHttpVct(vct string) bool {
 	return strings.HasPrefix(vct, "https://") || strings.HasPrefix(vct, "http://")
 }
 
+// RevocationChecker reports whether a stored credential instance is currently
+// revoked. The disclosure planner depends only on this narrow verb, keeping the
+// Token Status List mechanics out of this package (see services.RevocationService).
+type RevocationChecker interface {
+	IsRevoked(instance *models.IssuedCredentialInstance) bool
```
kamphuisem, line 53: Does this also check `suspended` state, because I think it should, as suspended credentials are also no valid candidates.

eudi/openid4vp/eudi_sdjwt_dcql/handler.go
```diff
@@ -149,6 +168,8 @@ func (h *SdJwtVcDcqlHandler) FindCandidates(query dcql.CredentialQuery) (*dcql.C
 			Attributes:                  attributes,
 			ExpiryDate:                  expiryUnix(batch),
 			Image:                       image,
+			Revoked:                     h.revocation != nil && h.revocation.IsRevoked(instance),
```
kamphuisem, line 171: Double-check; does each call to `h.revocation.IsRevoked(..)` update the entire status list?
As far as I can see, it potentially goes out and call/fetch externally.
Shouldn't we make a clear split between the fetching/caching/credential-updating job and remove the 'in-flight' fetching of status lists?

eudi/services/credential_service.go
```diff
@@ -283,12 +339,74 @@ func (s *credentialService) computeHashAndDeleteExisting(vc *sdjwtvc.VerifiedSdJ
...
-		instances[i] = models.IssuedCredentialInstance{
+		inst := models.IssuedCredentialInstance{
 			RawCredential: []byte(v.GetRawSdJwtVc()),
 		}
+		// Persist the status_list reference so the disclosure path and
+		// the refresh sweep can run without re-parsing the SD-JWT VC.
+		// At issuance time the holder verifier has just confirmed the
+		// bit reads StatusValid (or the credential has no status
+		// reference), so seed LastKnownStatus accordingly.
+		if v.IssuerSignedJwtPayload.Status != nil && v.IssuerSignedJwtPayload.Status.StatusList != nil {
```
kamphuisem, line 399: Double-check; this is stored per-instance, only for easy access during revocation checking/updating, right?

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Centralize ad-hoc `http.Client` instantiations into a single shared `common.HTTPClient`, giving one source of truth for outbound client configuration.
 
 ### Added
+- Support for the IETF OAuth Token Status List (draft-ietf-oauth-status-list-15) on SD-JWT VC credentials
```
kamphuisem: Maybe shorten the changelog changes, just to mention that we now support the IETF Token Status List spec, or something to that effect.

client/client.go
w-ensink: Potentially a "force refresh" option could be added in the app in the future, which would then use this function

eudi/credentials/statuslist/reference.go
w-ensink: Good question, we should think about that at some point

eudi/openid4vp/eudi_sdjwt_dcql/handler.go
w-ensink: This is a tough one that I also considered. It now only updates the status of the requested credential. If the update fails it'll fall back to the cached value. A live revocation check is potentially a good thing, but we might also get away with the startup refresh being done before the session gets to this point anyway. So it might be cleaner to only rely on cached values indeed.

eudi/credentials/statuslist/reference.go
w-ensink: For now we'll treat it as revoked

eudi/openid4vp/eudi_sdjwt_dcql/handler.go
w-ensink: We moved to a cache only check to simplify

eudi/jwt/key_providers.go
w-ensink: On `master`, `sdjwtvc/verify.go` used a
package-local `SdJwtKeyProvider`. This PR **removed** that type and replaced it
with `eudi/jwt.JwtKeyProvider` — the shared `x5c`-or-`kid`+DID dispatcher now used
by _both_ SD-JWT VC verification (`verify.go:720`) and Status List Token
verification (`statuslist/verifier.go:72`). The only difference between the two
call sites is the allowed `typ` list they pass in (`vc+sd-jwt` vs
`statuslist+jwt`). `eudi/jwt` is the separate package you're suggesting.
Reply: "Already merged — `JwtKeyProvider` in `eudi/jwt` replaces the old
`SdJwtKeyProvider`; both verifiers share it, differing only in the accepted
`typ`. `grep SdJwtKeyProvider` is now empty."

eudi/services/credential_service.go
w-ensink: Correct. `buildInstances` (credential_service.go:387) persists `StatusListURI` +
`StatusListIdx` on each instance so both the disclosure path (`IsRevoked`) and
the refresh sweep run without re-parsing the SD-JWT VC. `LastKnownStatus` is
seeded to `StatusValid` at issuance because the holder verifier has just
confirmed the live bit. The idx is genuinely per-instance (each one-time-use copy
has its own bit position — `validateStatusReferences` even enforces that copies
don't share an idx), so this can't be lifted to the batch.

kamphuisem (approved):

rubenhensen (approved):

DibranMulder (approved):

PR #581, opened 2026-06-16
Fix Dependabot security advisories, bump Go to 1.26, enforce go fix in CI

rubenhensen:
## Summary

Resolves all six open Dependabot alerts, raises the Go version to 1.26, and adds a `go fix` modernization gate to CI.

### Security: dependency updates

| Package | From | To | Advisories | Severity |
|---|---|---|---|---|
| `github.com/jackc/pgx/v5` | 5.5.5 | 5.10.0 | GHSA-9jj7-4m8r-rfcm, GHSA-j88v-2chj-qfwx | critical / low |
| `github.com/sirupsen/logrus` | 1.9.0 | 1.9.4 | GHSA-4f99-4q7p-p3gh | high |
| `golang.org/x/crypto` | 0.40.0 | 0.53.0 | GHSA-f6x5-jh6r-wrfv, GHSA-j5w8-q4qc-rx2x | medium (the vulnerable `ssh`/`ssh/agent` packages are used in `internal/common`) |
| `filippo.io/edwards25519` | 1.1.0 | 1.2.0 | GHSA-fw7p-63qq-7hpr | low (transitive via the MySQL driver) |

### Go 1.26

The patched `pgx/v5` (>= 5.9.0) requires Go 1.25, so the `go` directive in `go.mod` is raised. I went to **1.26** to also pick up the new `go fix` modernizers. CI already uses `go-version-file: go.mod` and the `Dockerfile` uses `golang:1-alpine`, so both follow automatically.

### `go fix` modernization + CI gate

- Ran `go fix ./...` across the codebase: `interface{}` → `any`, classic counter loops → `for range`/`min`/`max`, manual loops → `slices`/`maps` helpers, `&literal` → `new(expr)`, and removal of obsolete `// +build` lines.
- Removed the small helper functions (`stringPtr`, `strPtr`, `boolPtr`, `intPtr`, `uintPtr`) that `go fix` orphaned by inlining them into `new(expr)` — otherwise `staticcheck` (U1000) would flag them as unused.
- Added `go fix -diff ./...` to the lint job. It prints a diff and exits non-zero when a modernization fix is unapplied, keeping the tree at the `go fix` fixpoint.

The changes are organized into three commits (deps, `go fix`, CI) for easier review.

## Verification

Locally verified the full lint chain that CI runs, plus the unit tests that don't need the docker-compose service stack:

- `go build ./...` (incl. the cgo `sqlcipher` package) — ✅
- `go vet ./...` — ✅
[...]

PR #583, opened 2026-06-17
ci: drop windows/arm (removed in Go 1.26) and catch broken build targets on PRs

rubenhensen:
## Summary

The `Delivery` workflow started failing on `master` after #581 raised Go to 1.26:

```
go: unsupported GOOS/GOARCH pair windows/arm
```

Go 1.26 **removed the 32-bit `windows/arm` port** ([release notes](https://go.dev/doc/go1.26#windows)). `go tool dist list` under go1.26.3 confirms the only remaining Windows ports are `windows/386`, `windows/amd64`, and `windows/arm64`.

### Why PR checks didn't catch this

Two reasons:

1. The `Delivery` workflow (which actually cross-compiles each target via `.github/actions/build`, setting `GOOS`/`GOARCH`) only runs on `push: master` / `release` / `workflow_dispatch` — **not on `pull_request`**. So the failing job never ran while #581 was open.
2. `status-checks.yml` *did* have a `build` job with the same `os × arch` matrix that runs on PRs, but its step ran plain `go build ./...` with **no `GOOS`/`GOARCH`** — so all 9 legs compiled for the host (linux/amd64) and only labelled the uploaded artifact differently. The `build (windows, arm)` leg never cross-compiled, so it was green.

### Changes

- **delivery.yml**: exclude `windows/arm` from the release matrix.
- **status-checks.yml**: add a `build-release-artifact` job that cross-compiles every release target on PRs using the same `.github/actions/build` composite action Delivery uses (with the `windows/arm` exclude), so an unsupported target pair fails on the PR instead of after merge. The existing `build` job stays a single host build (cgo/sqlcipher) that still produces the `yivi-linux-amd64` binary the integration tests consume.
- **actions/build**: point the shared action at `go-version-file: go.mod` instead of a stale `^1.21` pin, so its toolchain matches the rest of CI.

### Impact

The `yivi-windows-arm.exe` (32-bit ARM Windows) release artifact is no longer produced. Windows-on-ARM is covered by `windows/arm64`; 32-bit ARM Windows is not a real consumer target.

Verified locally with go1.26.3 (matching CI) that `windows/arm` is the only matrix pair the toolchain no longer supports; all other targets (`./yivi`, `CGO_ENABLED=0`) cross-compile cleanly.

> [!NOTE]
> Once merged, consider adding `build-release-artifact` to the required status checks in branch protection so the cross-compile gate is enforced.

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -19,7 +19,7 @@ runs:
     - name: Set up Go
       uses: actions/setup-go@v5
       with:
-        go-version: ^1.21
+        go-version-file: go.mod
 
     - name: Determine artifact output filename
       id: artifact-name-generator
--- a/.github/workflows/delivery.yml
+++ b/.github/workflows/delivery.yml
@@ -140,6 +140,9 @@ jobs:
             arch: 386
           - os: darwin
             arch: arm
+          # Go 1.26 removed the 32-bit windows/arm port (windows/arm64 remains).
+          - os: windows
+            arch: arm
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/status-checks.yml
+++ b/.github/workflows/status-checks.yml
@@ -13,16 +13,9 @@ on:
 
 jobs:
 
+  # Host build of the whole module (cgo/sqlcipher enabled) that both compile-checks
+  # every package and produces the yivi-linux-amd64 binary the integration tests run.
   build:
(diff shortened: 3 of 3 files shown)


w-ensink (approved):

rubenhensen:
/dobby if you think this pr is approved, please actually approve the pr

DibranMulder (approved):

PR #613, opened 2026-07-11
Add mDoc age-verification implementation with device binding, certificate/credential validity checks, and test suite

khancak:
## Summary

Minimal ISO 18013-5 mDoc issuer → holder → verifier flow for the EU Age
Verification Blueprint (`eu.europa.ec.av.1`), with selective disclosure,
device binding, and a two-level certificate chain.

## What's included

- Issuer signs claims into a COSE_Sign1 MSO (`issuerAuth`), chained through
  a DS cert → self-signed IACA root, with the holder's public key embedded
  in `deviceKeyInfo`
- Holder does selective disclosure and signs a fresh `deviceAuth` per session
- Verifier checks the X.509 chain, MSO signature, per-item digests,
  `deviceAuth` against the embedded device key, **and the MSO's own
  `validityInfo` window separately from cert expiry**
- `COSEKey` fixed to use real CBOR integer keys (RFC 9053)
- `decode/` — standalone CLI for inspecting COSE_Sign1/CBOR blobs

## Tests

Full round trip, tamper/replay/clone rejection, untrusted-root rejection,
expired/not-yet-valid rejection for **both** certs and MSO validity
(via injectable clock, with cert `NotBefore` backdated 5 min so the two
checks can be tested independently), and a COSE-encoding regression test.

## Known gaps

No `DeviceRequest`/`DeviceResponse` wrapper, no transport/session
encryption, no verifier-side cert chain. Not production code.


testdata/mdoc/mdoc_full_spec.svg
DibranMulder, line 1: Please leave svg files in the wiki repo

DibranMulder (commented):
Please move code towards `eudi/credentials/mdoc` and also move the tests there. The `testdata` folder is only for test data and not for the actual tests.

So try to refactor code into:
- mdoc/issuer.go
- mdoc/issuer_test.go
- mdoc/verifier.go
- mdoc/verifier_test.go

PR #623, opened 2026-07-18
fix(eudi): make holder-binding key-metadata PK primary (Postgres upsert)

DibranMulder:
## Problem

On a **Postgres-backed** EUDI holder storage, every OpenID4VCI credential
redemption fails at the point of storing the holder binding key:

```
failed to store holder binding keys:
ERROR: ON CONFLICT DO UPDATE requires inference specification or constraint name (SQLSTATE 42601)
```

`ECDSAKeyMetadata` and `RSAKeyMetadata` tag `HolderBindingKeyID` as
`gorm:"uniqueIndex"`, so the models have **no primary key** — even though the doc
comment on both states *"KeyID is both the PK and FK to holderbindingkeys.id."*
When GORM upserts the key-metadata association while storing a `HolderBindingKey`,
it builds the `ON CONFLICT` target from the model's primary key; with none
defined it emits `ON CONFLICT DO UPDATE` with no target. SQLite/SQLCipher
tolerates that, but Postgres rejects it.

## Change

Tag `HolderBindingKeyID` as `gorm:"primaryKey"` on both `ECDSAKeyMetadata` and
`RSAKeyMetadata`, so GORM emits `ON CONFLICT ("holder_binding_key_id") DO UPDATE`
— the schema the comment already intends.

## Backwards compatibility

**No data migration is required; the change is backwards-compatible.**

- **Existing tables keep working.** `AutoMigrate` is additive — it does not drop
  the index the old `uniqueIndex` tag created. Both SQLite and Postgres resolve
  an `ON CONFLICT (holder_binding_key_id)` arbiter from **any** unique index on
  the column, not only a declared PRIMARY KEY. So the new upsert SQL resolves
  against an already-created table's residual unique index — no rebuild, no data
  loss.
- **Fresh tables** (any backend) are created with `holder_binding_key_id` as the
  formal PRIMARY KEY.
- **Verified**: on a Postgres schema, `AutoMigrate` re-runs cleanly with the new
  model (no error promoting/leaving the key column), and the previously-failing
  key store now succeeds.

**Known limitation (not blocking):** `AutoMigrate` does not *promote* an existing
table's unique index to a PRIMARY KEY (it doesn't alter PKs — SQLite can't
`ALTER ADD PRIMARY KEY` at all). So installs created before this change keep the
unique-index shape rather than a formal PK — functionally identical for the
upsert. Reconciling that would need a one-off table rebuild, which belongs with
the migration-framework work (`storage.go` already carries a `TODO` to separate
migrations from storage init). At present the EUDI holder storage ships no
versioned-migration framework and is not in a tagged release, so there are no
production SQLite deployments to reconcile — a dev wallet DB is simply recreated.

## Context

Surfaced running the holder OpenID4VCI flow against `eudi/storage` on
`gorm.io/driver/postgres` (server-side wallet) — the same "irmago assumed SQLite"
class of gap as the CGO-free storage split (#622).


DibranMulder:
**Compatibility / migration:** no data migration required. `AutoMigrate` is additive, so the index the old `uniqueIndex` tag created is not dropped, and both SQLite and Postgres infer the `ON CONFLICT (holder_binding_key_id)` target from any unique index on the column — so the new upsert resolves against existing tables without a rebuild or data loss. Fresh tables get the formal PK. `AutoMigrate` won't retroactively promote an existing unique index to a PRIMARY KEY, so pre-existing installs keep the unique-index shape (functionally identical for the upsert). Given the EUDI holder storage is unreleased and uses plain AutoMigrate (no versioned-migration framework yet), there are no production SQLite deployments to reconcile.

w-ensink (approved):

w-ensink (approved):

kamphuisem (approved):

PR #624, opened 2026-07-20
feat(eudi): WSCA-ready holder-key binding seams for OpenID4VCI/OpenID4VP

DibranMulder:
## Summary

Adds the seams needed to bind SD-JWT VC **holder keys to an external secure device (WSCA/HSM)** for both OpenID4VCI issuance and OpenID4VP presentation, so the holder private key never has to live in the wallet process. The seams are opt-in — when unset, behaviour is identical to today's software keys.

This PR is library-only: the standalone wallet facade and `sdjwtvc-wallet` CLI that originally accompanied these seams have been dropped (they were an unused proof-of-concept). The reusable seams remain and are consumed by the external `wallet-provider` WSCA adapter.

## What's included

**`eudi/holderkeys` (new, CGO-free package)**
- `HolderSigner` / `SoftwareHolderSigner` — the holder-key signing abstraction and its software implementation.
- `NewSignerKeyBinder` — a KB-JWT `sdjwtvc.KeyBinder` bridge that signs the KB-JWT through any external signer.
- Kept free of the sqlcipher (cgo) dependency so a WSCA adapter or a server-side (Postgres) holder can implement and wire it while building `CGO_ENABLED=0`.

**Pluggable seams in the existing `eudi/*` stack**
- `eudi/openid4vci`: `NewClient` takes a required `HolderKeyBinder`; callers pass `services.NewHolderBindingKeyService(...)` for the default software binder or a WSCA-backed implementation. (Explicit dependency injection rather than a hidden default — per review feedback.)
- `eudi/credentials/proofs`: `BuildWithES256Signer` signs the OpenID4VCI proof of possession via an external signer; the existing `Build(privKey)` is untouched.
- `eudi/openid4vp/eudi_sdjwt_dcql`: `NewSdJwtVcDcqlHandler` accepts an optional `sdjwtvc.KeyBinder` (variadic, backward-compatible).

**Docs**: `docs/wsca-holder-binding-integration.md` (WSCA holder-binding integration design).

## Verification
- Builds under cgo and `CGO_ENABLED=0` static (release + Docker); `./yivi` carries no sqlcipher/cgo dependency.
- `go vet`, `go fix -diff`, staticcheck (CI check set), and the new unit tests pass.
- Unit tests prove a KB-JWT and an OpenID4VCI proof signed through an external signer verify against the holder public key.
- A WSCA-backed implementation of these seams (in the `wallet-provider` repo) has been driven end-to-end against a live WSCA + SoftHSM.

## Notes
- Rebased onto current `master`, so this builds on top of the holder key-metadata PK fix (#623) and the CGO-free storage split (#622); it no longer carries them.
- `eudi/holderkeys` and `proofs.BuildWithES256Signer` have no in-repo caller — they are library API for the external WSCA consumer.
- Inherits existing `eudi/*` limitations (deferred issuance, DPoP, `did:jwk` KB-JWT verification, revocation, mdoc/W3C) documented in the design doc.


eudi/openid4vci/client.go
```diff
@@ -31,23 +31,32 @@ type Client struct {
 	currentSession *session
 	holderVerifier *sdjwtvc.HolderVerificationProcessor
 
+	// holderKeyBinder, when set, overrides the default software holder key
+	// binder used during issuance (e.g. a WSCA-backed binder).
```
w-ensink: I think it would be better to make this a mandatory param when instantiating the client. Instead of overriding a default just pass in the default at the places it was used before.

eudi/openid4vp/eudi_sdjwt_dcql/handler.go
```diff
@@ -61,16 +61,30 @@ type SdJwtVcDcqlHandler struct {
...
+// WSCA/HSM. When omitted, the default software binder over the eudi storage is
+// used, preserving existing behavior.
 func NewSdJwtVcDcqlHandler(
 	eudiStorage storage.Storage,
 	vctFetcher typemetadata.VctFetcher,
 	issuerFetcher typemetadata.IssuerFetcher,
+	keyBinder ...sdjwtvc.KeyBinder,
 ) *SdJwtVcDcqlHandler {
-	keyService := services.NewHolderBindingKeyService(eudiStorage.Db())
+	var binder sdjwtvc.KeyBinder
```
w-ensink: I think the `keyBinder` parameter should be mandatory here and then `sdjwtvc.NewDefaultKeyBinder(services.NewHolderBindingKeyService(eudiStorage.Db()))` should be passed in wherever it's used currently.

w-ensink (approved):

PR #637, opened 2026-07-24
Fix randomblind attribute names, gated on protocol 2.9 (non-breaking alternative to #151)

rubenhensen:
Non-breaking alternative to #151, for discussion.

## The bug (same as #151)

`CredentialType.RandomBlindAttributeNames()` reports which attributes of a credential type are random blind. It did so by taking the metadata-offset indices from `RandomBlindAttributeIndices()` (which are `i+1`, because the cryptographic `AttributeList` keeps a metadata attribute at index 0) and matching them against the **0-based** `AttributeTypes` slice. Two index spaces, matched as if they were one, so it returned the neighbouring attribute's identifier — or, when the random blind attribute is declared last (as `votingnumber` is in `stempas`), nothing at all.

The value is put on the wire in the credential request (`randomblindIDs`) and compared by the client against its own recomputation, and it shows up in server logs. Verification still passed because both peers ran the identical buggy code, so their equal-but-wrong lists compared equal. The cryptography was never affected — issuance uses `RandomBlindAttributeIndices()`, which is correct and untouched.

## Why not just fix it (#151)

#151 fixes the function and puts the corrected identifiers straight into the existing wire field. Because old clients compare that field with frozen exact-equality, a patched server (`["votingnumber"]`) and an unpatched app (`[]`) then disagree and the client aborts with *"mismatch in randomblind attributes between server/client"*. So #151 is a breaking change for random blind issuance across mismatched versions.

## This approach: gate the fix on protocol 2.9

The consistency check is redundant with a crypto agreement that already holds (both peers derive the random blind **indices** identically), so a cross-version abort is a false alarm. That gives room to gate the corrected behaviour on the negotiated protocol version:

- `RandomBlindAttributeNames()` is fixed (reads the `RandomBlind` flag directly).
- `RandomBlindAttributeNamesLegacy()` reproduces the old off-by-one output, kept for wire-compat with peers below 2.9.
- The client-side check moves from `CredentialRequest.Validate` (no version in scope) to `IssuanceRequest.GetCredentialInfoList` (has the negotiated version). It compares against the corrected names at ≥2.9 and the legacy names below 2.9.
- The server downgrades the identifiers it sends to a `<2.9` client in `handleGetClientRequest`, right after version negotiation.

Compatibility, once 2.9 is enabled:

| Server | Client | Negotiated | Wire value | Client compares with | Result |
|---|---|---|---|---|---|
| new | old | 2.8 | legacy | legacy (frozen) | pass |
| old | new | 2.8 | legacy | legacy (`Below(2,9)`) | pass |
| new | new | 2.9 | correct | correct | pass + real check |

## Important: `maxProtocolVersion` stays at 2.8

Protocol 2.9 already exists but is only half-landed: the client advertises `9` in `supportedVersions` and `keyshare.go` gates a "new keyshare protocol version" on `Below(2, 9)`, but the server's `maxProtocolVersion` is pinned at `2.8` so 2.9 never negotiates. Because the version counter is linear, bumping the server past 2.8 would also switch on that pending keyshare change for every current app.

So this PR deliberately does **not** bump `maxProtocolVersion`. It lands the machinery and corrects server-side logging now (the stored request is stamped with the correct names at session creation); the wire behaviour and the strict check activate automatically when 2.9 is enabled. Until then, behaviour on the wire is byte-for-byte identical to today for all negotiated versions.

## Changes

- `irma/descriptions.go` — fix `RandomBlindAttributeNames`; add `RandomBlindAttributeNamesLegacy`.
- `irma/requests.go` — remove the wire comparison from `CredentialRequest.Validate`; add version-gated `IssuanceRequest.checkRandomBlindConsistency`, called from `GetCredentialInfoList`.
- `irma/server/irmaserver/handle.go` — downgrade the sent identifiers for `<2.9` clients after negotiation.
- test + CHANGELOG.

`go build ./...`, `go vet ./...`, and the blind issuance session tests (`TestRequestorServer/BlindIssuanceSession`, `TestIrmaServer/BlindIssuanceSessionDifferentAmountOfRandomBlinds`) pass.


rubenhensen:
Thanks for the sharp review — both addressed in e7a3dc3.

**Blocking (log-render path).** You're right that `GetCredentialInfoList` is shared with the read-only log path (`LogEntry.GetIssuedCredentials` → `logs.go:154`), where running the check could panic on a nil `LogEntry.Version` and could make a historical log fail to render after a config change. I removed the check from `GetCredentialInfoList` entirely and moved it to the live client session path (`irmaclient/session.go`, right before the `GetCredentialInfoList` call in `processSessionInfo`). The method is now exported as `IssuanceRequest.CheckRandomBlindConsistency` and treats a nil version as pre-2.9 (`version != nil && !version.Below(2, 9)`), so the log path never touches it. Added a regression test that renders an issuance request with a populated `CredentialInfoList` and a nil version — it no longer panics and skips the check.

**Nit (in-place mutation).** Also fixed. Instead of mutating `session.Rrequest.SessionRequest()`, the `<2.9` downgrade now rewrites the copy returned by `getRequest` / `getClientRequest` (via a small `downgradeRandomBlindNames` helper applied at the three serve sites: the `<2.8` branch, the `ClientSessionRequest` branch, and `handleSessionGetRequest`). The stored request keeps the corrected names stamped at `api.go`, so the "corrects server-side logging" claim holds.

Tests: the four-case version-gate matrix and the end-to-end 2.9 issuance still pass, plus the new `irma`-package regression tests (log-render nil-version, and the client-side gate across 2.8 / 2.9 / nil). `go build`, `go vet`, `gofmt` clean.

/dobby ready for another look.

PR #653, opened 2026-07-30
feat(client): tell the app when a credential's revocation status changes

w-ensink:
## What

For an SD-JWT VC, the wallet never told the app about a revocation. It wrote the
new status to the database and stopped there, so a credential revoked while the
app sat idle stayed rendered as valid until the user happened to do something that
re-read the list. For an idemix credential the signal did exist — `Revoked(cred)`
— but it repeated on every rediscovery.

This adds the missing signal, and takes the opportunity to give `client` its own
app-facing callback interface instead of borrowing `irmaclient`'s.

- **`client.ClientHandler`** — `client.New` now takes this instead of
  `irmaclient.ClientHandler`. It carries only what the app acts on:
  `CredentialsChanged`, `ReportError`, and the enrollment and change-pin results.
  An unexported `irmaHandler` adapter implements the `irmaclient` surface and
  forwards, so `IrmaClient`'s vocabulary stops leaking into the app.
- **`CredentialsChanged()`** — one signal for "the credentials you are showing
  are out of date", replacing `UpdateAttributes` and `Revoked(cred)`. It now
  fires for revocations too, on both credential technologies.

## Why one callback and not a revocation-specific one

The first cut had a separate `RevocationStatusUpdated()`. It was collapsed
because the app re-requests the whole credential list either way — a second
callback carried no information the app acted on differently. `Revoked(cred)`'s
credential identity is dropped for the same reason; `IrmaClient` still fires it
internally, so re-widening later is a one-line change in the adapter if a
per-credential notification is ever wanted.

The trade-off worth knowing: the app can no longer tell *why* its credentials
changed. Issuance, deletion, revocation, and a logo download are one signal.

## Only a change fires it

A signal on every check would have the app re-reading its whole credential list
on a timer, forever. Both paths report transitions, not state:

- **SD-JWT VC.** `ListInstancesWithStatusReference` now also selects
  `last_known_status`, so the sweep can compare. `RefreshStatuses` returns
  `(changed int, err error)` — the number of *batches* whose status moved, since
  only one representative per batch is ever checked — and `client.Client` decides
  that `> 0` is worth waking the app for. A re-confirmation, and a sweep whose
  fetches all failed, both count zero.
- **Idemix.** `IrmaClient` reports revocation as *state*: a revoked credential's
  witness never advances (gabi returns `ErrorRevoked` before touching it), so its
  staleness only grows and the periodic witness-update job rediscovers the same
  revocation every few tens of seconds. The adapter reports each credential once.

## No irmaclient behaviour changes

`irma/irmaclient` is untouched apart from a counter on its test mock. The idemix
signal is derived entirely from the existing `Revoked(cred)` callback, which
already fires on exactly the right event — the adapter maps it and dedupes it.

## Breaking — irmamobile must ship in lockstep

`client.New` takes a `client.ClientHandler`. The app must:

- rename `UpdateAttributes` → `CredentialsChanged`;
- fold whatever it did on `Revoked(cred)` into `CredentialsChanged` (it no longer
  receives which credential);
- drop `UpdateConfiguration`.

`Client.RefreshStatuses(ctx) error` keeps its signature.

Internal, no app impact: `services.RevocationService.RefreshStatuses` returns
`(changed int, err error)`, and `db.CredentialStatusInstance` carries
`LastKnownStatus`.

## Tests

- `client/status_change_notification_test.go` — the wiring, against a real
  `Client` with a seeded status-bearing batch and a real status server: a change
  wakes the app, a re-confirmation does not, a known revocation is not
  re-reported, and a caller who cancels gets no callback. Plus a unit test that
  the adapter reports each idemix revocation once.
- `internal/sessiontest/revocation_test.go` — `TestRevocation*/NotifiesApp`,
  a real idemix revocation at a revocation-enabled server, driven through
  `client.Client`. Needs the docker-compose postgres, like its 22 siblings.
- `internal/sessiontest/openid4vci_statuslist_test.go` — the SD-JWT counterpart,
  a real credential revoked at the status-list agent. In the opt-in status-list
  group (needs the full docker stack), verified by hand.
- `eudi/services/credential_status_refresh_test.go` — `changed` counts: zero on
  re-confirmation, zero on failure, one per batch rather than per row.

Every notification path was mutation-checked — each assertion fails when the
code under it is disabled.

## Follow-up

`Client.Close` can still close storage while a background job is running against
it, because no wallet job can be cancelled. A `jobCtx` fixing this for the status
sweep alone was written and reverted — it buys an inconsistent contract for one of
five job families. Doing it properly means covering all five, and picking up two
existing TODOs (`InitJobs`, `IrmaClient.Close`) and the fact that
`common.HTTPClient` has no timeout along the way.


w-ensink:
/dobby fix your comments and check again until you consider it good enough. Then approve the PR.

rubenhensen (approved):

PR #663, opened 2026-07-31
Add mDoc (ISO 18013-5) support to issuance and presentation

khancak:
Add mDoc (ISO 18013-5) support to issuance and presentation

mso_mdoc credentials can now be issued over OpenID4VCI and presented over OpenID4VP, alongside SD-JWT VC. Previously eudi/credentials/mdoc existed only as a prototype with its own duplicate protocol layer, unreachable from any production path.

Approach

Rather than adding a parallel mdoc protocol stack, this reuses the existing generic ones and introduces a single format seam on the issuance side, mirroring dcql.DcqlCredentialQueryHandler on the presentation side.

services.CredentialFormatParser (new) turns one raw credential from a CredentialResponse into a format-agnostic ParsedCredential. openid4vci/session.go now dispatches on credentialConfig.Format instead of hardcoding SD-JWT verification, so credential offer, token request, nonce endpoint, and proof-of-possession are shared rather than duplicated. NewClient derives the parser registry itself from the config and holder verifier it already receives — its signature is unchanged from master.

ParsedCredential carries an SdJwtVc *sdjwtvc.VerifiedSdJwtVc escape hatch, so every existing SD-JWT path (VCT integrity, cnf key matching, attribute flattening) operates on exactly the type it always did. hashForSdJwtVc and matchHolderBindingKey are untouched, so stored SD-JWT hashes stay stable and dedup is unaffected.

Presentation adds eudi/openid4vp/mdoc_dcql, built to match eudi_sdjwt_dcql. Three batch helpers (IsBatchValid, BatchExpiryUnix, BatchInstanceCountRemaining) moved from eudi_sdjwt_dcql into dcql/batch_display.go, bodies moved verbatim. mdoc_dcql does not call them — MdocBatch and SdJwtVcBatch share no fields to write one implementation over without generics, so mdoc keeps its own equivalents (services.MdocBatchIsValid and two private helpers in mdoc_dcql/handler.go). The move is about where the code can be reached from, not about the two formats sharing it: dcql has no import of services, where eudi_sdjwt_dcql does, so this is the only place a services caller could reach them without a cycle, if one needs to.

Storage

Each credential format owns its storage end to end. mso_mdoc lives in its own tables — mdoc_batches, mdoc_batch_instances, mdoc_device_keys — with its own models (models.MdocBatch, models.MdocBatchInstance, models.MdocDeviceKey), stores (db.MdocStore, db.MdocDeviceKeyStore), key minter (services.MdocKeyService) and persistence path (services.NewMdocCredentialService).

The two formats deliberately do not share a table. They evolve on different specifications, on different timelines, and towards different hardware seams, so one table would make every schema change to one format a migration of the other's rows, and every format-specific column a nullable one the other format has to remember to ignore.

That lets the mdoc columns be named after ISO 18013-5 rather than after SD-JWT VC: the MSO's three validityInfo timestamps, none of them nullable because 18013-5 makes all three mandatory; a typed namespaces cache, so DCQL matching and display do not re-decode CBOR; one credential_issuer column, since an mdoc has no iss claim and the document signer certificate travels inside every instance's IssuerSigned bytes rather than in a column; and the OpenID4VCI display metadata as two JSON snapshots rather than the normalized five-table tree. An mdoc device key is stored and looked up by its JWK thumbprint whatever binding method the proof used, which retires the DID-derivation workaround the shared key table forced on mdoc.

The SD-JWT VC Go types are renamed to say what they are — models.CredentialBatch → models.SdJwtVcBatch, models.IssuedCredentialInstance → models.SdJwtVcBatchInstance, db.CredentialStore → db.SdJwtVcStore — while their deployed table and column names stay: credential_batches and issued_credential_instances, pinned by TableName() and explicit foreignKey tags. SdJwtVcBatch.ProcessedSdJwtPayload keeps its name for the same reason, and with the split it is accurate again — it holds SD-JWT payloads only. AutoMigrate is the only schema mechanism here and cannot rename a column: it would ADD the new one, which SQLite rejects for a NOT NULL column once the table has rows, so every wallet holding a credential would fail to open its database. credentials_schema_test.go pins the deployed column name, and TestAutoMigrateAddsIssuerVerifiedToPopulatedBatchTable runs the upgrade an installed wallet actually performs: a legacy schema without the column, populated with a credential, then migrated.

Developer note: no mdoc data is migrated, because none has shipped. A developer or staging wallet built from an earlier state of this branch holds its mdoc batches in the SD-JWT table and must be wiped before use, or those rows show up as broken SD-JWT credentials.

Unrelated fixes found along the way

Both are independent of mdoc and reviewable on their own:

- internal/common/common.go — SaveFile used path.Dir where it needed filepath.Dir. On Windows path.Dir doesn't understand \, so it returned "." and every "atomic" write put its temp file in the working directory rather than the destination. This is production code, not test-only. It also fixed two long-standing test failures (TestDcqlCandidateSelection, TestInstantiateClientWithExistingIrmaStorage) and stopped test runs littering the repo with orphaned temp files.
- GetCredentialMetadataList — pre-sized a slice with make(..., len(m)) then appended into it, returning a nil entry per credential and panicking downstream.


khancak:
/dobby yo d dawg, do this pr

khancak:
/dobby i aint got all day lil bro...

khancak:
/dobby dobs review the current version of the pr

eudi/credentials/mdoc/holder.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+// ============================================================
+// HOLDER
+// ============================================================
+
+// Holder represents the wallet app on the user's device
+// deviceKey is generated locally — private key never leaves the device (TEE/Secure Enclave in production)
+// Only the PUBLIC key is sent to the issuer at issuance time
+type Holder struct {
```
w-ensink: Might want to make Holder an interface so we can create an implementation that uses the Secure Enclave/TPE.

eudi/openid4vp/mdoc_dcql/handler.go
```diff
@@ -0,0 +1,647 @@
...
+		if err := cbor.Unmarshal(instance.RawCredential, &doc); err != nil {
+			return nil, fmt.Errorf("decode stored mdoc: %w", err)
+		}
+
+		disclosed, err := selectiveDiscloseByPaths(&doc, sel.ClaimPaths)
+		if err != nil {
+			return nil, fmt.Errorf("selective disclosure: %w", err)
+		}
+
+		privKey, err := decodeECDSAPrivateKey(instance.HolderBindingKey.PrivateKey)
```
w-ensink: I don't think the private key should be required here in memory. It should go via a signing interface, so that it can also be done using a TPM

eudi/openid4vp/mdoc_dcql/handler.go
```diff
@@ -0,0 +1,647 @@
...
+	if err := json.Unmarshal(batch.ProcessedSdJwtPayload, &resolved); err != nil {
+		return nil, err
+	}
+	return resolved, nil
+}
+
+// mdocPathParts splits a DCQL claim path into its mandatory [namespace,
+// elementIdentifier] components. mso_mdoc claim paths are always exactly two
+// string components deep -- ISO 18013-5 has no nested claims, unlike SD-JWT.
+func mdocPathParts(path []any) (namespace, elementIdentifier string, ok bool) {
```
w-ensink: Probably better to make this an error instead of a boolean, so we can be more descriptive about what's wrong with the DCQL query

internal/sessiontest/eudi_pid_python_issuer_mdoc_test.go
```diff
@@ -0,0 +1,655 @@
...
+	eudiPidIssuerPyAvDocType            = "eu.europa.ec.av.1"
+)
+
+// testSessionHandlerForEudiPidPythonIssuerMdoc is registered alongside the
+// SD-JWT subtests in session_handler_test.go.
+func testSessionHandlerForEudiPidPythonIssuerMdoc(t *testing.T) {
+	t.Run("issues age verification mdoc", testEudiPidPythonIssuerIssuesAvMdoc)
+	t.Run("discloses issued mdoc to EUDI Kotlin verifier", testEudiPidPythonIssuerDisclosesAvMdoc)
+	t.Run("batch instances are spent one per presentation", testEudiPidPythonIssuerBatchIsSingleUse)
+	t.Run("discloses two claims at once", testEudiPidPythonIssuerDisclosesTwoClaims)
```
w-ensink: Most of these seem like disclosure tests instead of issuance tests

internal/sessiontest/eudi_pid_python_issuer_mdoc_test.go
```diff
@@ -0,0 +1,655 @@
...
+
+	session = awaitSessionState(t, sessionHandler)
+	requireSessionState(t, session, sessionId, clientmodels.Type_Issuance, clientmodels.Status_RequestPermission)
+	require.Len(t, session.OfferedCredentials, 1)
+
+	// The offer the user is asked to accept names the docType out of the signed
+	// MSO. It has to come from the credential: an mso_mdoc credential
+	// configuration publishes no vct, so a dialog built from the issuer's
+	// advertised metadata alone would name nothing at all.
+	require.Equal(t, eudiPidIssuerPyAvDocType, session.OfferedCredentials[0].CredentialId,
```
w-ensink, line 547: Would be good to also assert the expected attributes, their human readable names and their values here, as well as a potential image and the issuer metadata. Ideally in at least two locales, so we can determine that translations also work correctly.

internal/sessiontest/eudi_pid_python_issuer_mdoc_test.go
```diff
@@ -0,0 +1,655 @@
...
+		"the issuer advertises batch_credential_issuance, so a batch of one means the wallet never asked for more than a single attestation")
+	require.Equal(t, batch.BatchSize, batch.RemainingCount,
+		"a freshly issued batch has spent nothing yet")
+
+	// Every instance must carry its own device key: one reused key across the
+	// batch would correlate the presentations that the batch exists to separate.
+	// Read straight from the database rather than off the batch: the display
+	// preloads GetBatchesByDocType uses do not include instances.
+	var instances []models.IssuedCredentialInstance
+	require.NoError(t, c.EudiStorageForTesting().Db().
```
w-ensink: Never access the database directly in one of these tests. It leaks implementation details into the tests, making them fragile. Instead only use the `client.Client` public interface.

internal/sessiontest/eudi_pid_python_issuer_mdoc_test.go
```diff
@@ -0,0 +1,655 @@
...
+			Proceed:         true,
+			TransactionCode: &offer.TxCode,
+		},
+	})
+
+	// The wallet may refuse at the permission step or after fetching the
+	// credential, depending on how far it gets before verifying; either way it
+	// must end in an error and store nothing.
+	session = awaitSessionState(t, sessionHandler)
+	if session.Status == clientmodels.Status_RequestPermission {
```
w-ensink: This if-statement should be removed. We need this to be 100% deterministic.

internal/sessiontest/openid4vp_mdoc_av_disclosure_test.go
```diff
@@ -0,0 +1,783 @@
...
+	testSession, requestJwt := startMdocAvSessionCapturingRequest(t, c, 1, sessionHandler,
+		verifierHost, createMdocAvAuthRequestRequest(t, issuer))
+
+	session := testSession.ClientSession
+	requireSessionState(t, session, 1, clientmodels.Type_Disclosure, clientmodels.Status_RequestPermission)
+	require.Equal(t, clientmodels.Protocol_OpenID4VP, session.Protocol)
+
+	// No display metadata was stored with the credential (the AV profile does not
+	// specify any), so the credential name falls back to the raw docType and the
+	// claim has no label. The claim path is the two-component
```
w-ensink: Where should we get the metadata from? Should e inline a hardcoded version of all known credential metadata or something? Either way the metadata should be reported to the frontend.

internal/sessiontest/openid4vp_mdoc_av_disclosure_test.go
```diff
@@ -0,0 +1,783 @@
...
+	// [namespace, elementIdentifier] form mdoc matching requires.
+	requireDisclosurePlan(t, session.DisclosurePlan, expectedDisclosurePlan{
+		Choices: []expectedPickOneChoice{
+			{
+				Owned: []expectedPlanCredential{{
+					CredentialId: avDocType,
+					Name:         avDocType,
+					IssuerName:   "",
```
w-ensink: Issuer name is empty, that should not be the case. Probably some implementation missing?

internal/sessiontest/openid4vp_mdoc_av_disclosure_test.go
```diff
@@ -0,0 +1,783 @@
...
+// the wallet in a state a real issuer will not hand out on demand — an expired
+// credential, or a batch whose instances are all spent.
+type avSeedOptions struct {
+	BatchSize      uint
+	RemainingCount uint
+	// ExpiresIn is relative to now, and may be negative for an expired batch.
+	ExpiresIn time.Duration
+}
+
+func seedAvMdocBatch(t *testing.T, c *client.Client, opts avSeedOptions) *stdmdoc.Issuer {
```
w-ensink: Remove this, only use real issuance with an issuance server and openid4vci to get credentials into the wallet for the test. Using the database like this exposes implementation details and doesn't simulate a real world scenario.

yivi/cli/eudicli/internal/localstack/localstack.go
```diff
@@ -0,0 +1,376 @@
+// Package localstack drives the EUDI reference containers that docker-compose.yml
```
w-ensink: I'm not sure we want all of these CLI tools for Mdoc, but it's an interesting proposal

kamphuisem (dismissed):

common/clientmodels/types.go
```diff
@@ -185,6 +211,21 @@ type Credential struct {
...
+	// nothing marking it. A client that ships its own labels for credential types
+	// it knows uses this to substitute one only where the issuer left a gap,
+	// rather than overriding what the issuer did publish. Always true when the
+	// issuer published no display metadata at all.
+	//
+	// Set for EUDI credentials, whose text comes from issuer metadata. False for
+	// IRMA credentials: their text comes from the Yivi scheme, which carries every
+	// language it supports, so a client has nothing to substitute and no gap to
+	// fill.
+	DisplayIsFallback bool `json:"display_is_fallback"`
```
w-ensink: I don't think this is necessary

common/clientmodels/types.go
```diff
@@ -291,6 +332,21 @@ type SelectableCredentialInstance struct {
...
+	// than on an empty string, so a Dutch wallet is handed real English text with
+	// nothing marking it. A client that ships its own labels for credential types
+	// it knows uses this to substitute one only where the issuer left a gap,
+	// rather than overriding what the issuer did publish. Always true when the
+	// issuer published no display metadata at all.
+	//
+	// Set for EUDI credentials, whose text comes from issuer metadata. False for
+	// IRMA credentials: their text comes from the Yivi scheme, which carries every
+	// language it supports, so a client has nothing to substitute and no gap to
+	// fill.
```
w-ensink: Remove this

.github/workflows/status-checks.yml
```diff
@@ -126,7 +126,20 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    # 30, above the 18m `go test -timeout` in docker-compose.yml's test service, so
+    # the two caps cannot fight: a suite that genuinely overruns should be killed by
+    # go test — which names the tests still running — rather than by the job, which
+    # only reports that the runner was cut off.
```
w-ensink: too much comment

eudi/credentials/mdoc/crypto.go
```diff
@@ -0,0 +1,378 @@
...
+// last. That is not merely untidy — every structure here is either covered by a
+// signature or used to check one, so two parties can decode the same bytes to
+// different documents while both believe they parsed correctly. An MSO carrying
+// two docType entries, or two digests under one digestID, is exactly the shape
+// of a parser-differential attack. Rejecting is the only reading that makes the
+// decode deterministic.
+//
+// The clause's companion rule about definite lengths is deliberately not
+// enforced; see indefLengthMode.
+var mdocDecMode = mustDecMode(cbor.DecOptions{
```
w-ensink, line 35: consider making this a function instead. Global variables is not a great idea usually.

eudi/credentials/mdoc/crypto.go
```diff
@@ -0,0 +1,378 @@
...
+// round-trips against this codebase — decoding would use the same wrong
+// mapping — while failing against every spec-compliant verifier. Worse, the
+// encoding is covered by the signed MSO digest, so an mdoc issued with the
+// wrong one cannot be repaired after the fact.
+//
+//	1  = kty  (key type: 2 = EC2)
+//	-1 = crv  (curve:    see coseCurves)
+//	-2 = x    (x coordinate, the curve's field size in bytes)
+//	-3 = y    (y coordinate, the curve's field size in bytes)
+type COSEKey struct {
```
w-ensink: Are there any libraries for COSE so we don't have to define and maintain this ourselves?

eudi/credentials/mdoc/holder.go
```diff
@@ -0,0 +1,205 @@
...
+// Holder is the wallet app on the user's device, reduced to the only two
+// operations that need the device key: handing out the public half at issuance,
+// and signing a DeviceAuthentication at presentation.
+//
+// It is an interface so the private half never has to exist in this process.
+// DefaultHolder is the software implementation used by tests and by the current
+// wallet; an implementation backed by StrongBox, TrustZone or the Secure Enclave
+// satisfies the same two methods without the key ever being extractable. See
+// NewHolderFromSigner for the cheapest route to one.
+type Holder interface {
```
w-ensink, line 32: Should we rename this to KeyBinder or should we stay closer to the mdoc specific naming?

w-ensink, line 32: Is this for testing only? Which implementation is used in production? If it's only for testing, we should name it accordingly and consider making it private or moving it to a test file.

eudi/credentials/mdoc/holder.go
```diff
@@ -0,0 +1,205 @@
...
+// go-cose — cose.NewSigner(AlgorithmES256, aP384Key) succeeds — so the wrong
+// pairing produces a signature that transmits fine and fails at the verifier
+// with nothing naming the cause.
+//
+// Ed25519 and Ed448 are absent for the reason given on coseCurves: an
+// ed25519.PublicKey is not an *ecdsa.PublicKey, and admitting one means widening
+// this type and the device-key binder.
+func deviceAuthAlgorithmFor(curve elliptic.Curve) (cose.Algorithm, error) {
+	switch curve {
+	case elliptic.P256():
```
w-ensink, line 124: Is this the same as the hardcoded strings we use in `digestFuncFor()` in `crypto.go`?

eudi/credentials/mdoc/issuer.go
```diff
@@ -0,0 +1,349 @@
...
+//   - Phase 3: iacacert itself is signed by the EU AV Blueprint root CA
+type Issuer struct {
+	iacakey  *ecdsa.PrivateKey
+	iacacert *x509.Certificate
+
+	dskey  *ecdsa.PrivateKey
+	dscert *x509.Certificate
+}
+
+func NewIssuer() (*Issuer, error) {
```
w-ensink: This seems like a test only thing. Name it accordingly. Having a hardcoded certificate in a function with a name that suggests it's ready for production is not a good idea.

eudi/credentials/mdoc/mdoc.go
```diff
@@ -0,0 +1,245 @@
...
+	// CBOR null under a present namespace key, which is neither an array nor an
+	// absent namespace, and no conformant verifier can read it.
+	//
+	// Erroring rather than dropping the namespace, for the reason
+	// selectiveDiscloseByPaths gives: a presentation quietly missing what the user
+	// consented to disclose is indistinguishable downstream from a verifier that
+	// asked for less. Unreachable through the DCQL handler, which refuses an empty
+	// claim set up front — this is the guard for direct callers of the exported
+	// function.
+	if len(disclosed) == 0 {
```
w-ensink, line 233: What if the user chose to disclose nothing?

eudi/credentials/mdoc/profile.go
```diff
@@ -0,0 +1,245 @@
...
+// Verification profile, which uses the same string for both.
+//
+// Source: the AV Blueprint's architecture and technical specification
+// (ageverification.dev). Note this is the mdoc docType, not the OpenID4VCI
+// credential configuration id — the reference issuer advertises the latter as
+// "eu.europa.ec.eudi.age_verification_mdoc" while minting documents whose
+// docType is this. Profile dispatch keys off the docType inside the signed MSO,
+// so the configuration id never reaches here.
+const (
+	AgeVerificationDocType   = "eu.europa.ec.av.1"
```
w-ensink, line 39: Should we hardcode this or can this be removed?

eudi/credentials/mdoc/profile.go
```diff
@@ -0,0 +1,245 @@
...
+	// validityPeriod is how long an issued credential stays valid. 18013-5 sets
+	// no bound; the AV Blueprint recommends "a maximum period of three (3) months
+	// from the date of issuance".
+	validityPeriod time.Duration
+}
+
+// profileFor returns the rules that apply to docType. An unrecognised docType
+// gets general ISO/IEC 18013-5 and nothing more, which is the whole point: this
+// package targets mso_mdoc, and AV is one profile on top of it.
+func profileFor(docType string) documentProfile {
```
w-ensink, line 102: Not sure if we should have any special treatment for the age verification profile. Making sure the correct attributes are issued seems more like the responsibility of the issuer to me.

eudi/credentials/mdoc/README.md
```diff
@@ -0,0 +1,737 @@
+# mDoc Issuer → Holder → Verifier (Go)
+
+An implementation of ISO 18013-5 mDoc selective disclosure, built against the EU Age
```
w-ensink, line 3: This document should be removed.

eudi/credentials/mdoc/verifier.go
```diff
@@ -0,0 +1,1163 @@
...
+	x509.ExtKeyUsageNetscapeServerGatedCrypto:      "netscapeServerGatedCrypto",
+	x509.ExtKeyUsageMicrosoftCommercialCodeSigning: "microsoftCommercialCodeSigning",
+	x509.ExtKeyUsageMicrosoftKernelCodeSigning:     "microsoftKernelCodeSigning",
+}
+
+// Verifier holds the pre-installed trust anchor (IACA root cert)
+// Phase 1: our own test self-signed IACA root
+// Phase 2: Yivi's own IACA root, manually distributed to verifiers
+// Phase 3: EU AV Blueprint root CA (from official AP trust list)
+type Verifier struct {
```
w-ensink, line 248: Is this used in the app or only in tests?

eudi/credentials/mdoc/verifier.go
```diff
@@ -0,0 +1,1163 @@
...
+// than checking a single caller-supplied namespace, it verifies digests for
+// every namespace actually present in mdoc.IssuerSigned.NameSpaces. Needed
+// because at issuance the caller doesn't know the docType's namespace set
+// ahead of a single Verify(mdoc, namespace) call, unlike presentation (which
+// already knows the namespace from the DCQL claim path being disclosed).
+// Returns the decoded namespace -> elementIdentifier -> value map alongside
+// the same VerificationResult shape Verify returns (Attributes is left at
+// its zero value here — namespace-scoped attributes are the return map, not
+// the result, since a single flat map would lose the namespace boundary).
+func (v *Verifier) VerifyAllDisclosedNamespaces(mdoc *MDoc) (map[string]map[string]any, VerificationResult) {
```
w-ensink: Maybe good to create a type (alias) out of `map[string]map[string]any` so it's easier to see what it represents.

eudi/credentials/mdoc/verifier.go
```diff
@@ -0,0 +1,1163 @@
...
+	// SignDeviceAuth detaches it before returning, matching the AV
+	// Blueprint spec's own example (deviceSignature payload: null).
+	deviceMsg, err := decodeCoseSign1(deviceAuthBytes)
+	if err != nil {
+		result.Valid = false
+		result.Error = fmt.Sprintf("decode deviceAuth cose: %v", err)
+		return result
+	}
+
+	// Rebuild the DeviceAuthentication payload the holder signed. Two of its four
```
w-ensink, line 1008: Too much comments

eudi/credentials/mdoc/verifier.go
```diff
@@ -0,0 +1,1163 @@
...
+
+// decodeDeviceNameSpaces validates that raw really is
+// `DeviceNameSpacesBytes = #6.24(bstr .cbor DeviceNameSpaces)` and returns the
+// namespaces it wraps, their contents left undecoded.
+//
+// The emptiness of the result is what the profile check tests, rather than a byte
+// comparison against tag24Wrap(map[string]any{}): CBOR admits more than one
+// encoding of an empty map, and comparing bytes would reject a conformant holder
+// for choosing a different one — the very brittleness this replaced.
+func decodeDeviceNameSpaces(raw cbor.RawMessage) (map[string]map[string]cbor.RawMessage, error) {
```
w-ensink: Maybe good to make a type (alias) out of `map[string]map[string]cbor.RawMessage`

eudi/credentials/mdoc/verifier_salt_credential_test.go
```diff
@@ -0,0 +1,312 @@
...
+	"github.com/veraison/go-cose"
+)
+
+// Salt-length rejection, exercised through the entry points a wallet actually
+// calls, on complete credentials an issuer could really have sent.
+//
+// verifier_salt_test.go already covers the comparison itself: it calls
+// verifyNamespaceDigests directly with a hand-built item and a matching digest.
+// That pins the check but cannot show it is reachable — an earlier step in
+// Verify or VerifyAllDisclosedNamespaces could reject or accept the credential
```
w-ensink: Too much comments

eudi/credentials/mdoc/verifier_test.go
```diff
@@ -0,0 +1,1274 @@
...
+// the MSO's window relative to a test clock.
+func issueWithValidity(t *testing.T, issuer *Issuer, validFrom, validUntil time.Time) *MDoc {
+	t.Helper()
+
+	holder, err := NewHolder()
+	if err != nil {
+		t.Fatalf("NewHolder: %v", err)
+	}
+	deviceKey, err := coseKeyFromECDSA(holder.PublicKey())
+	if err != nil {
```
w-ensink: use the require package for this instead of this pattern:
```
if err != nil {
		t.Fatalf("NewHolder: %v", err)
}
```

eudi/credentials/mdoc/verifier_salt_test.go
```diff
@@ -0,0 +1,82 @@
...
+	item := IssuerSignedItem{
+		DigestID:          1,
+		Random:            make([]byte, minSaltLength-1),
+		ElementIdentifier: "age_over_18",
+		ElementValue:      true,
+	}
+	items, digests := wrapItemWithMatchingDigest(t, item)
+
+	attrs, err := verifyNamespaceDigests(items, digests, sha256Digest)
+	if err == nil {
```
w-ensink: replace by require package

eudi/openid4vp/dcql/dcql_handler.go
```diff
@@ -23,15 +23,72 @@ func NewDcqlHandler(handlers []DcqlCredentialQueryHandler) *DcqlHandler {
 type DcqlResult struct {
 	// Per-query results keyed by credential query ID.
 	QueryResults map[string]*CredentialQueryResult
-	// Maps credential hashes to their DCQL query IDs.
-	HashToQueryId map[string]string
+}
+
+// CandidateQuery names the DCQL credential query that one owned candidate of a
+// pick-one answers, alongside the claim paths that candidate would disclose.
+type CandidateQuery struct {
```
w-ensink, line 42: I don't understand this one

DEMO.md
```diff
@@ -0,0 +1,229 @@
+# Running the mDoc age-verification demo on a phone
```
w-ensink: Remove this file

yivi/cli/eudicli/mint-session/localstack.go
```diff
@@ -0,0 +1,481 @@
+// This file drives the EUDI reference containers that docker-compose.yml brings
+// up: asking the Python issuer for a credential offer, starting a presentation at
```
w-ensink: Remove the mint-session package

testdata/eudi/verifier/verifier_scheme_data.json
```diff
@@ -39,6 +39,148 @@
 					"family_name",
 					"birthdate"
 				]
+			},
```
w-ensink, line 42: We can also start using wildcards here or completely remove the authorization mechanism

testdata/eudi-pid-issuer-py/patches/dynamic_func.py
```diff
@@ -0,0 +1,343 @@
...
+        attributes_req = {}
+        attributes_req2 = {}
+        issuer_claims = {}
+
+        for namescape in namescapes:
+            attributes_req.update(attributes_by_namespace[namescape]["mandatory"])
+            attributes_req2.update(attributes_by_namespace[namescape]["optional"])
+            issuer_claims.update(attributes_by_namespace[namescape]["issuer"])
+
+    else:  # "dc+sd-jwt"
```
w-ensink, line 138: What is this file for?

client/client.go
```diff
@@ -151,7 +152,15 @@ func New(
 	}, statusListCache)
 	revocationService := services.NewRevocationService(statusChecker, credStore)
 
-	credentialService := services.NewCredentialService(credStore, hbkStore, eudiStorage.FileSystem(), revocationService, currentLocale)
+	// Rewrite any credential hash still computed without the issuer. Runs before
+	// the wallet can be asked about its credentials, because a stale hash makes a
```
w-ensink, line 156: why do we need to migrate the credential hashes?

common/clientmodels/enums.go
```diff
@@ -20,9 +20,18 @@ const (
 )
 
 // CredentialFormat identifies the format of a credential.
+//
+// Adding a constant here is a breaking change for the Yivi app until irmamobile
+// is updated to match. irmamobile mirrors this list as the Dart enum
```
w-ensink, line 25: Remove this comment

eudi/credentials/mdoc/holder.go
w-ensink: Checked: production, not test-only.

`DefaultHolder` is what the wallet uses today. The device private key lives in this process, read from storage on demand — the same arrangement `sdjwt.DefaultKeyBinder` has on the SD-JWT side. Non-test callers:

- `eudi/services/mdoc_device_key_binder.go:20,64`
- `eudi/openid4vp/mdoc_dcql/handler.go:33`

The interface exists for what replaces it: a StrongBox / TrustZone / Secure Enclave key satisfies the same two methods without ever being extractable, via `NewHolderFromSigner` — a platform key handle only has to implement `Public` and `Sign`.

The doc comment is what invited this question: it said "used by tests and by the current wallet", which reads as hedging about which is which. Rewritten to state plainly that this is the production implementation and that the interface is the hardware seam.

eudi/credentials/mdoc/holder.go
w-ensink: No — two different IANA registries, answering two different questions, sharing no strings:

|  | `digestFuncFor` (crypto.go:163) | `deviceAuthAlgorithmFor` (holder.go:118) |
|---|---|---|
| Input | `MSO.digestAlgorithm`, a string off the wire | `elliptic.Curve`, a type |
| Output | a hash function | `cose.Algorithm` |
| Values | `SHA-256` / `SHA-384` / `SHA-512` | `ES256` / `ES384` / `ES512` |
| Clause | 18013-5 9.1.2.5, Table 21 | 18013-5 9.1.3.6 |

The first resolves how the issuer says it computed the digests; the second resolves which signature algorithm the spec pairs with a given curve. No overlap, so nothing to consolidate.

eudi/credentials/mdoc/verifier.go
w-ensink: Production, on every disclosure — reached from `eudi/openid4vp/mdoc_dcql` and `eudi/services`.

Worth recording why `verificationOptions` and `revocationLists` are functions rather than pools captured at construction: `Configuration.Reload` rebuilds the wallet's trust models (toggling developer mode adds and drops the staging anchors) and *replaces* the pools instead of mutating them. Anything captured at construction therefore goes stale in both directions — it misses anchors that were added, and keeps honouring anchors that were dropped.

testdata/eudi-pid-issuer-py/patches/dynamic_func.py
w-ensink: It is live test infrastructure: a patched copy of the EU reference PID issuer's `app/dynamic_func.py`, bind-mounted over the image's own at `docker-compose.yml:372`.

Stock upstream, `populate_pdata` builds the credential from the claims declared in the issuer's credential configuration and copies a value out of the offer only where one is named. An element the offer asks for and the configuration does not declare is dropped before the MSO is signed — issuance still returns HTTP 200, and the element is simply absent with no error at either end. The patch lifts that for the age-verification namespace, because ISO 18013-5 and the AV profile both leave the set of `age_over_NN` thresholds open, so a test needs to mint one the configuration does not list.

Nothing in the file said any of that — it carried only the upstream EC licence header. Fixed: added `testdata/eudi-pid-issuer-py/patches/README.md` (upstream origin, what changed, why, and how to make an edit take effect) and a `PATCHED COPY` header on the file itself. That README is also where the explanation from the now-deleted `DEMO.md` lives.

client/client.go
w-ensink: Because this PR changes the hash. On `master`:

```go
func hashForSdJwtVc(credType string, processedSdJwtPayloadBytes []byte) (string, error)
```

on this branch:

```go
func hashForSdJwtVc(credType, issuerIdentifier string, processedSdJwtPayloadBytes []byte) (string, error)
```

The issuer became part of the content identity, so the same claims from a different issuer are now a different credential.

Without the migration, a wallet that already holds SD-JWT credentials computes the new-style hash on re-issuance, finds no stored batch under it, and stores a second copy: the user sees the credential twice and the old row is never matched again. The recomputation is exact rather than a guess (every input the hash takes is already a column) and idempotent, which is why it is safe to run on every startup.

Open question though, and the one actually worth deciding: does the issuer belong in the hash *in an mdoc PR* at all? If that change is separable from mdoc support, it and its migration would be a cleaner standalone PR.

w-ensink: What is the ID for an issuer and can it change over time?

w-ensink: We use the `iss` field for SD-JWTs and the `SAN` of the issuer certificate. According to ISO 18013-5 the certificate is required to have that field.

eudi/credentials/mdoc/mdoc.go
w-ensink: Mdoc is required to disclose at least one issuer signed item, so this if statement is correct.

eudi/credentials/mdoc/agility_test.go
```diff
@@ -0,0 +1,274 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+	cose "github.com/veraison/go-cose"
+)
+
+// ============================================================
+// ALGORITHM AND CURVE AGILITY
+//
+// 9.1.2.4: "For verifying the signature, the mdoc reader shall support all of
```
kamphuisem: Not sure if this is quoted from the standard, but probably better to change this.

eudi/credentials/mdoc/crypto.go
```diff
@@ -0,0 +1,372 @@
...
+// The clause's companion rule about definite lengths is deliberately not
+// enforced; see indefLengthMode.
+var mdocDecMode = mustDecMode(cbor.DecOptions{
+	DupMapKey:   cbor.DupMapKeyEnforcedAPF,
+	IndefLength: indefLengthMode,
+})
+
+// indefLengthMode is the one setting in mdocDecMode left lenient.
+//
+// 8.1's "indefinite-length items shall be made into definite-length items" is
```
kamphuisem: Not sure if this is quoted from the standard, but probably better to change this.

eudi/credentials/mdoc/crypto.go
```diff
@@ -0,0 +1,372 @@
...
+func tag24WrapBytes(innerBytes []byte) ([]byte, error) {
+	tagged := cbor.RawTag{
+		Number:  24, // IANA registered tag: embedded CBOR
+		Content: cbor.RawMessage(mustMarshal(innerBytes)),
+	}
+	return cbor.Marshal(tagged)
+}
+
+// tag24Unwrap decodes Tag-24 wrapped CBOR bytes into T — the inverse of
+// tag24Wrap/tag24WrapWithMode. Tag 24 means "this byte string contains a
```
kamphuisem: Not sure if this is quoted from the standard, but probably better to change this.

kamphuisem (approved):

w-ensink (approved):

PR #721, opened 2026-09-09
Refactor to use github.com/lestrrat-go/jwx

### LINKED ISSUE #480: Replace golang-jwt with lestrrat-go/jwx
For the EUDI implementation, we needed JWK and JWE stuff etc, which we found in the `lestrrat-go/jwx` lib.
This causes some developer confusion, because we have existing JWT library `golang-jwt` in use for other JWT related code.

We'd like to unify all code to use the `lestrrat-go/jwx` library for JWT related code, as this offers more code we need.

lestrrat:
Fixes #480

Author of jwx here. I was just researching jwx usage from various GitHub projects, and stumbled upon this.
I don't know if you want it or not, but here it is.

While this PR does replace golang-jwt/jwt with jwx, the way it's being used -- mainly through maps isn't as idiomatic jwx-y as it could be. I opted to keep the seams mostly as they were to not introduce gratuitous changes, but if you're going all the way in, you could certainly just do away with all the `map[string]any` and just pass `jwt.Token` or a wrapper thereof around.

Full disclosure, this PR was created mainly using Claude Code, with steering from yours truly.
If this helps, cool. If not, feel free to close.
Thanks!


kamphuisem:
Thanks for your input @lestrrat and wonderful to see you taking time to participate in our project, awesome!

We were aware we needed to do more changes regarding the migration (and update to jwx v4), but your input is very much appreciated! We'll take a better look at it soon.

lestrrat:
I'll check the CI results

lestrrat:
pushed a lint commit. test and CodeQL failures are probably something that already existed and/or flake?

PR #722, opened 2026-09-11
Proximity Flow Implementation for mDoc

khancak:


khancak:
> Reviewed and posted a CHANGES_REQUESTED review on the PR with 5 inline comments: [#722 (review)](https://github.com/privacybydesign/irmago/pull/722#pullrequestreview-5178565233)
> 
> Bugs flagged:
> 
> * A device-key leak when re-issuance is refused
> * A reader-side gap that can't verify deviceMac-authenticated responses
> * A single-use mdoc instance getting burned even when a later step in the same multi-credential disclosure fails
> * An unlocked read letting two concurrent presentations reserve the same single-use instance
> * A proximity-session error path that neither wipes session key material nor distinguishes a harmless late message from a genuine fault
> 
> Comment style/density looked consistent with the rest of the codebase, so no notes there. Take a look when you get a chance!

PR #728, opened 2026-09-21
Enforce certificate bound attribute authorization in OpenID4VP

### LINKED ISSUE #727: OpenID4VP: client_metadata bypasses the attribute authorization bound to a Yivi issued verifier certificate
## Summary

In `RequestorCertificateStoreVerifierValidator.ParseAndVerifyAuthorizationRequest`, the attribute level authorization check that is bound to a Yivi issued relying party certificate is only reached when the authorization request does **not** carry `client_metadata.client_name`. A verifier that holds any Yivi issued certificate can therefore skip the check — and present itself under a name of its choosing — simply by including `client_metadata` in its request.

## Details

The verifier's identity is resolved in `eudi/openid4vp/x509_verifier_validator.go` with this chain:

1. `client_metadata.client_name` from the authorization request, if present
2. the scheme data in the certificate, if it is a Yivi issued certificate
3. the certificate CN as a fallback

`queryValidator.ValidateCredentialQueries` lives inside branch 2 only. Because branch 1 is an `if` and branch 2 an `else if`, a request containing `client_metadata.client_name` never enters branch 2, so:

- the `authorized` attribute sets in the certificate are not enforced, and the verifier can request any credential and any attribute;
- `requestorInfo.Organization.LegalName` and the logo come from the request instead of from the certificate.

The request is still fully authenticated: the JWT signature, the certificate chain, the CRL check and the `x509_san_dns` / `x509_hash` binding all pass normally. Consequently `Verified` is `true` for such a request (`eudi/openid4vp/client.go`), so the wallet presents it with the "known party" styling under the self-asserted name.

## Impact

Any holder of a Yivi issued verifier certificate can request credentials and attributes beyond what its certificate authorizes, and can have the consent screen display an organisation name and logo it chose itself. The embedded disclosure policy carried in the certificate — the control that limits what each relying party may ask for — is not enforced on that path.

Certificates issued by third parties are unaffected in substance: they carry no scheme data, so there is no certificate bound policy to enforce for them either way.

## Fix

Consult the certificate first. When it is a Yivi issued certificate it is authoritative for the verifier's identity and its authorized attribute sets, so `ValidateCredentialQueries` always runs and `client_metadata` cannot override the name or logo. `client_metadata` is used only for certificates that carry no scheme data, where no certificate bound policy exists.

## Traceability

Found during the gap analysis against the OWASP MAS Profile for EU Digital Identity Wallets (Security Requirements V1.0), tracked there as MAS-069.

- Requirement #69 — "The app uses privacy-preserving functionality" (MASWE-0069), the embedded disclosure policy asset
- Requirement #11 — "The app properly verifies cryptographic signatures" (MASWE-0011), relying party authentication

DibranMulder:
Fixes #727.

## Problem

`RequestorCertificateStoreVerifierValidator.ParseAndVerifyAuthorizationRequest` resolved the verifier's identity from `client_metadata.client_name` first, and only from the certificate's scheme data when that was absent. `queryValidator.ValidateCredentialQueries` sat in the certificate branch alone, so an authorization request carrying `client_metadata` never reached it.

A verifier holding any Yivi issued certificate could therefore request credentials and attributes its certificate does not authorize, and have the consent screen present it under an organisation name and logo of its own choosing. The request is otherwise fully authenticated — signature, chain, CRL and `x509_san_dns` / `x509_hash` binding all pass — so it was shown with the "known party" styling.

## Change

The certificate is consulted first. When it carries Yivi scheme data it is authoritative for both the verifier's identity and its authorized attribute sets: the queried credentials are always validated, and `client_metadata` can no longer override the legal name or the logo.

`client_metadata` is still used for certificates issued by third parties. Those carry no scheme data, so there is no certificate bound policy to enforce and their behaviour is unchanged.

## Compatibility

A Yivi issued certificate whose `authorized` attribute sets do not cover what the verifier actually requests will now be rejected where it previously passed, if that verifier was sending `client_metadata`. That is the intended effect, but it is worth checking against production verifier certificates before release: a relying party whose certificate is out of date with its request will start failing.

Verifiers that relied on `client_metadata` purely for display will now be shown under the name and logo in their certificate.

## Follow-up (not in this PR)

`Verified` is set to `endEntityCert != nil` in `eudi/openid4vp/client.go`, so a third party certificate that supplies its own `client_metadata` name is still presented as a verified party. Deciding the policy for non-Yivi certificates is tracked separately in the TODO that this code already carries.


DibranMulder:
Superseded by #729.

This PR was written against v1.3.1, the version irmamobile pins, and I missed that 1.4.0 had already landed on master. 1.4.0 closes the attribute authorization half of #727 by lifting `ValidateCredentialQueries` out of the identity chain, with test coverage, so that part of this PR is redundant.

#729 rebases onto master and keeps only what is still open: the display name and logo of a Yivi issued certificate can still be overridden by self-asserted `client_metadata`.

